PR #14, opened 2020-01-14
update import

creanme:
import CommonJS modules with star () import:* as something


prevwong:
The tsconfig in this project has `esModuleInterop` setting enabled, and I actually prefer the default import syntax that we currently use as I find it more readable. Apart from being a different syntax, I don't think changing to the star imports adds much value, if not any. 

Regardless, thank you for the PR and I really appreciate it! 👍🏼

PR #17, opened 2020-01-16
Added option to serialize JSON as a tree.

### LINKED ISSUE #13: Improve JSON documentation & output
I couldn't find any examples of what the JSON format looked like, and I wanted to be sure the format was relatively independent of Craft itself. However, the JSON uses node IDs instead of nesting the nodes themselves, which I find confusing.

For example, with these Frame contents:
```
<Frame>
  <Canvas is={CraftContainer}>
    <CraftComponent>
      <TextComponent text="my text here" />
    </CraftComponent>
  </Canvas>
</Frame>
```
The JSON output from `query.serialize()` is:
```
{
  "canvas-ROOT": {
    "type": {
      "resolvedName": "CraftContainer"
    },
    "isCanvas": true,
    "props": {},
    "displayName": "Canvas",
    "custom": {},
    "nodes": [
      "node-yxVcZGJj"
    ]
  },
  "node-yxVcZGJj": {
    "type": {
      "resolvedName": "CraftComponent"
    },
    "props": {
      "children": [
        {
          "type": {},
          "props": {
            "text": "text component"
          }
        }
      ]
    },
    "displayName": "CraftComponent",
    "custom": {},
    "parent": "canvas-ROOT",
    "_childCanvas": {
      "d02fab54-ddf2-4356-b79d-5660d54915df": "canvas-u-wv625gX"
    }
  },
  "canvas-u-wv625gX": {
    "type": {
      "resolvedName": "CraftContainer"
    },
    "isCanvas": true,
    "props": {},
    "displayName": "Canvas",
    "custom": {},
    "parent": "node-yxVcZGJj",
    "nodes": [
      "node-aKYJYqPZ0"
    ]
  },
  "node-aKYJYqPZ0": {
    "type": {},
    "props": {
      "text": "my text here"
    },
    "displayName": "TextComponent",
    "custom": {},
    "parent": "canvas-u-wv625gX"
  }
}
```
I would expect something more like:
```
[
  {
    "type": "CraftContainer",
    "nodes": [
      {
        "type": "CraftComponent",
        "nodes": [
          {
            "type": "TextComponent",
            "props": {
              "text": "my text here"
            }
          }
        ]
      }
    ]
  }
]
```
It would also be nice to have some examples of implementing custom serializers, like [SlateJS](https://docs.slatejs.org/concepts/09-serializing) does. I'm currently using SlateJS as a WYSIWYG React editor, but I'd love to switch to Craft for the nicer drag & drop functionality.

Hyperkid123:
closes #13

Adds an option to serialize the JSON output as a tree instead of a flat object.

If the param is set to true, each node will get the `children` attribute. It as an array of objects representing the node children.

I did not remove any attributes so you have all the data you would need to convert it back to the flat object.

Sample: 

```JSON
{
  "type": {"resolvedName":"Container"},
  "isCanvas":true,
  "props":{"background":"#eeeeee","padding":5},
  "displayName":"Canvas",
  "custom":{},
  "nodes":["canvas-wYA65kzm"],
  "children":[{
    "type":{"resolvedName":"Container"},
    "isCanvas":true,
    "props":{"background":"#ffffff","padding":20},
    "displayName":"Canvas",
    "custom":{},
    "index":0,
    "parent":"canvas-ROOT",
    "nodes":["canvas-4ZraoNqU"],
    "children":[{
      "type":{"resolvedName":"Container"},
      "isCanvas":true,
      "props":{"background":"#ffffff","padding":20},
      "displayName":"Canvas",
      "custom":{},
      "index":0,
      "parent":"canvas-wYA65kzm",
      "nodes":["canvas-K37CJhds"],
      "children":[{
        "type":{"resolvedName":"Container"},
        "isCanvas":true,
        "props":{"background":"#ffffff","padding":20},
        "displayName":"Canvas",
        "custom":{},
        "index":0,
        "parent":"canvas-4ZraoNqU",
        "nodes":["canvas-vVWN6zCr"],
        "children":[{
          "type":{"resolvedName":"Container"},
          "isCanvas":true,
          "props":{"background":"#ffffff","padding":20},
          "displayName":"Canvas",
          "custom":{},
          "index":0,
          "parent":"canvas-K37CJhds",
          "nodes":["node-_2a8bFPV"],
          "children":[{
            "type":{"resolvedName":"Button"},
            "props":{"size":"small","variant":"contained","color":"primary","text":"Click me"},
            "displayName":"Button",
            "custom":{},
            "index":0,
            "parent":"canvas-vVWN6zCr"
          }]
        }]
      }]
    }]
  }]
}
```

### DIFF
--- a/packages/core/src/editor/query.tsx
+++ b/packages/core/src/editor/query.tsx
@@ -33,6 +33,19 @@ import { transformJSXToNode } from "../utils/transformJSX";
 const getNodeFromIdOrNode = (node: NodeId | Node, cb: (id: NodeId) => Node) =>
   typeof node === "string" ? cb(node) : node;
 
+function mapToTree(node, nodes) {
+  if (node && node.nodes && node.nodes.length > 0) {
+    if (!node.children) {
+      node.children = [];
+    }
+    node.nodes.forEach(id => {
+      node.children.push(mapToTree(nodes[id], nodes));
+    });
+  }
+
+  return node;
+}
+
 export function QueryMethods(Editor: EditorState) {
   const options = Editor && Editor.options;
 
@@ -64,8 +77,9 @@ export function QueryMethods(Editor: EditorState) {
     },
     /**
      * Retrieve the JSON representation of the editor's Nodes
+     * @param createTree output JSON will be returned as a tree instead of flat object.
      */
-    serialize(): string {
+    serialize(createTree: Boolean): string {
       const simplifiedNodes = Object.keys(Editor.nodes).reduce(
         (result: any, id: NodeId) => {
           const {
@@ -76,10 +90,13 @@ export function QueryMethods(Editor: EditorState) {
         },
         {}
       );
+      if (createTree) {
+        return JSON.stringify(
+          mapToTree(simplifiedNodes[ROOT_NODE], simplifiedNodes)
+        );
+      }
 
-      const json = JSON.stringify(simplifiedNodes);
-
-      return json;
+      return JSON.stringify(simplifiedNodes);
     },
     /**
      * Determine the best possible location to drop the source Node relative to the target Node
--- a/packages/docs/docs/api/useEditor.md
+++ b/packages/docs/docs/api/useEditor.md
@@ -46,7 +46,7 @@ const { connectors, actions, query, ...collected } = useEditor(collector);
     ]],
     ["query", "Query", [
       ["createNode", "(child: React.ReactElement) => Node", "Create a Node from a React element"],
-      ["serialize", "() => String", "Return the current Nodes in JSON"],
+      ["serialize", "(createTree: boolean) => String", "Return the current Nodes in JSON, When set to true, JSON will be returned as a nested tree instead of flat object."],
       ["getOptions", "() => Object", "Get the options specified in the &lt;Editor /&gt; component"],
       ["getDropPlaceholder", 
         "(sourceNodeId: NodeId, targetNodeId: NodeId, pos: {x: number, y: number}, nodesToDOM?: (node: Node) => HTMLElement = node => node.dom)",


prevwong:
Nice work, thanks for this! :100: 

I just have some concerns/comments:
1. Since we longer need to identify Nodes by their `id`, so I think it would be a little bit wasteful if we still kept the attributes like `id` and `nodes` around (maybe replace `nodes` with `children` ?).

2. It seems that your code only handles child `nodes`, but not `childCanvas`. 
```jsx
const HeroComp = () => {
  return (
    <div>
      <Canvas id="top">
         <h2>Hi</h2>
      </Canvas>
   </div>
  )
}
```
For the above component, we would have a `Node` that looks something like this:
```jsx
{
   "node-blah" : {
      ....
      _childCanvas: {
        "top" : "canvas-a"
     }
  },
  "canvas-a" : { ... }
}
```
Maybe something like this could work:
```jsx
{
   "node-blah" : {
      ....
      _childCanvas: {
        "top" : {
           // serialized canvas-a here
        }
     }
  },
  
}
```
3. If we are going with this structure, I think it's best we get rid of the old one. Because right now, if someone fed the JSON output that you proposed back into the editor, our deserializer would have to handle both the old and new structure. I prefer less variations in favor of maintainability. What do you think ?

Hyperkid123:
@prevwong thanks for the feedback. I do agree with what you have written, I just did not want to introduce anything disruptive without your approval.

> 1. Since we longer need to identify Nodes by their id, so I think it would be a little bit wasteful if we still kept the attributes like id and nodes around (maybe replace nodes with children ?).

That was my original thought. I think that `index`, `nodes` and `parent` attributes are no longer necessary. Is there any other you think I can remove/replace?

> 2. It seems that your code only handles child nodes, but not childCanvas.

Yes, you are right I did not consider this. I will fix that.

> 3. If we are going with this structure, I think it's best we get rid of the old one.

I personally prefer the tree structure. From my experience, it's easier to use trees for rendering the UI later on because it has the same structure as the code would have. So if you would like to expand the scope of this PR to switch just to tree structure, I am up for it.

One more thing I would think about is having the output just JSON object, not a string. Right now, if you want to do some post-processing, you have to parse it again. I think that a common use case would be getting the JSON, stripping most of the data related just to the editor (type, displayName, etc.) and renaming some of the attributes to match some rendering loop. And stringifying an object is not really an issue. What do you think about that?

prevwong:
> Is there any other you think I can remove/replace?

Let's remove `name`, `displayName`, `parent`, `index`, `isCanvas`. These values should be recoverable. I should've done this previously, whoops. :laughing:  

> childCanvas

Oh, just a heads up. Currently, it's still `_childCanvas` , but I was actually planning on renaming it to `childCanvas` for consistency.

> I personally prefer the tree structure. 

Just to clarify, we're still keeping the actual state in it's flat object form, correct ? The reason for doing this is that it would be much easier to access each Node by their `id` rather than looping through to find a match. 

> One more thing I would think about is having the output just JSON object, not a string. 

Good idea, let's go with this!

Hyperkid123:
> Just to clarify, we're still keeping the actual state in it's flat object form, correct ? The reason for doing this is that it would be much easier to access each Node by their id rather than looping through to find a match.

Yes, having the internal structure as a flat object is ok. I was referring only to the input/output format.

Hyperkid123:
@prevwong so I am looking at the canvas nodes and I have a question. Is it safe to assume, that nested (child) canvas is basically a virtual not which would technically not be rendered in the final product?

When I think about this purely as a consumer, I would not want any extra noise in my data. For example if I have would have the tree output like this:
```js
{
  name: "canvas-ROOT",
  children: [{
    name: 'node that has _childCanvas attribute',
    children: [{
      name: 'canvas-NESTED',
      children: [{
        name: 'some-Button',
      }, {
        name: 'some-Text'
      }]
    }]
  }]
}
```

The **node that has _childCanvas attribute** node is something that I would not probably want to see in my JSON output used for rendering the UI.

Now that said, I can now see that this is integral for the editor and necessary for the correct reproduction of the state.

This made me thinking. Maybe it's not the best idea to merge the editor schema and the rendering schema into one.

Maybe it would be better to keep them separate. Keep the `serialize` output as it is now, and either add another method to the query or create separate function and export it, which would create tree hierarchy and remove any and all attributes or nodes which don't have any use for rendering. 

What do you think?

prevwong:
If I understood your question correctly, the `childCanvas` **is** part of the final product, since it contains reference to an actual Node which contains information about what should be rendered in place of the Canvas inside a User Component.

```jsx
<Canvas> (i)
   <HeroComp /> (ii)
</Canvas>

const HeroComp = () => {
  return (
    <div>
      <Canvas id="top"> (iii)
         <h2>Hi</h2>
      </Canvas>
   </div>
  )
}
```
- For (i), (ii), we have our usual Nodes. But for (iii), there is a bit of a disconnect since it's nested inside a User Component. So, Craft solves this by creating a Node, and attaches it as a `childCanvas` in (ii). So every time (ii) re-renders, we can still ensure that the information in (iii) is left undisturbed.

ackvf:
Having two JSON output formats actually makes sense.

1. A full-feature deserializable JSON state which can reproduce the craft.js instance
2. A simplified JSON which can't be used as input in craft.js but can be used by a rendering engine

That being said, the 1. might as well stay flat since the internal state is flat for obvious reasons
and for 2. actually introduce a new standalone convertor which would be able to convert
state -> simplified JSON
full-feature JSON -> simplified JSON

What do you think?

prevwong:
Now that I think about the `childCanvas`. I'm not too sure how will this even work outside of Craft. 

```jsx
const HeroComp = () => {
  return (
    <div>
      // How would this Canvas be able to reconstruct it's saved state 
     // if the HeroComp is used outside of Craft ? 
      <Canvas id="top">
         <h2>Hi</h2>
      </Canvas>
   </div>
  )
}
```

This makes the usage of the JSON (simplified or not) in said custom rendering engine, outside of Craft, kinda meaningless. If you need Craft to just render the saved state without any editing features, you can do so by setting the `enabled` prop to false. As it stands right now, I can't think of any viable usage of the JSON output outside of Craft.  

I think that the simpler structure (the one I suggested to @Hyperkid123 which contains the `childCanvas`) would be more storage efficient than the flat JSON structure that we have now, since we would not need to store the `Node` ids and some other properties. Also, I think the simpler structure might be more logical when doing preprocessing or when you have to create the JSON yourself by hand. The only caveat I can think of is that the deserialising process will now have to be more complex.

ackvf:
> I can't think of any viable usage of the JSON output outside of Craft

Our usecase is this
1) devs create new blocks and prop editors
2) content administrators use craft.js UI to compose pages and store them as JSONs in database
3) customers visit a page which is rendered by our own render function from the stored JSON (without craft.js)

As mentioned, there can even be two JSON outputs, one simplified and efficient for the page rendering and the other with all metadata required for craft.js.

prevwong:
> customers visit a page which is rendered by our own render function from the stored JSON (without craft.js)

I've already explained disabling the editing features, but I still do not understand why do you need to build your own render function ? 

Using the JSON with your own rendering function is not viable because it's very likely that there will be User Components which has `<Canvas />` in them. 

Look at the Hero example again, a render function alone looking from the JSON data will not be able to somehow restore/inject the state into `<Canvas id="top />`

ackvf:
Yes, since Craft.js supports disabling editing, it's also an option, but still we don't want to load unnecessary libraries to the client, so a light-weight renderer could be better.

It could also be possible to work around Canvases if we knew, they are always rendered with div, for example, i.e. we would never use `<Canvas is={Component}/>` but instead use an empty `<Canvas />` within a User Component, which would simply have the purpose of a place-holder.
Or we could simply unwrap all the children inside a Canvas right to the Parent, so given this User Component

```
<A>
  <Canvas>
    // user moves <B/> here
    // user moves <C/> here
  </Canvas>
</A>
```

we would not render this

```
// whatever A and B renders
<A>
  <div>
    <B/>
    <B/>
  </div>
</A>
```

but this


```
// whatever A and B renders
<A>
  <B/>
  <B/>
</A>
```


This is what I meant by "introducing additional markup and attributes". These intermediary divs also might break already existing style sheets, when we convert our components into User Components for the Editor.

Hyperkid123:
@prevwong As @ackvf mention, sometimes it might be the best idea to use the editor as the renderer.

There might also be another use case of why you want to use different renderer entirely.

Take a form input for example. If you want to add input type 'hidden', you obviously don't want to show if in the UI, but you still need to somehow display it in the editor and let the user configure it.

Or you might have some design system independent tool for building the UI, and from that, the schema is distributed and used in different applications with different look and feel.

You might go even a step further a don't render the UI directly from the JSON, but generate actual code like html, jsx etc. And there you definitely don't want components like Canvas to "pollute" the end result.

As I look deeper into the code I am starting to think that maybe it's not the best idea to integrate some tree output into the component itself. At the end of the day, it's just a simple recursive function.
And I can imagine that everybody would require a slightly different output anyway.

Maybe adding just an example to docs would be enough? @prevwong @ackvf what do you think?

ackvf:
Yes exactly, since the renderer is really just a simple recursive function, it should be fairly easy to let users implement one that fits their specific use cases.

What we are really doing is we just traverse the JSON, look for component type, props and children and pass that to `React.createElement`.

My idea then was that whenever I encounter type Canvas, I will simply ignore the node and move directly to its children.

PS: Given how simple the renderer actually is, there could be one in the repo which users can extend on, right? Anyway, I still don't want to load the whole library on the end-client, but perhaps tree-shaking could solve this, when only the renderer is imported.

prevwong:
> What we are really doing is we just traverse the JSON, look for component type, props and children and pass that to React.createElement.

This is not enough. Like I said, my main concern is with the Canvas inside User Components. The problem is, when you render a component. You don't have control over what it renders. If the User Component has a `<Canvas />`, you can't just tell the `<Canvas />` what to render using a custom function.

Remember that the `<Canvas />` defines a Canvas Node which contains information such as the `type` to render as, and more importantly, the child `nodes` to render. The suggestion made by @ackvf only solves the `type` part, but not the latter.

Let's say we have the following JSON:
```jsx
const Hero = () => {
  return (
     <div>
        <h2>Hero</h2>
        <Canvas id="top">
           <h3>...</h3>
       </Canvas>
     </div>
  )
}

// For illustration, let's say our JSON is like this
{
   "node-a" : {
      ....
      type: Hero,
      _childCanvas: {
        "top" : {
		   isCanvas: true,
           type: "div",
           props: {},
           nodes: [
			{type: "p", props: {children : "I've been dropped here"}}
           ]
        }
     }
  },
}
```

Now let's just say Craft.js is not involved anymore and we are using your custom rendering function. When your custom rendering function is traversing the JSON and it reaches `node-a`, the type is `Hero`. 
- Ideally, you would want to render `Hero`. 
- But `Hero` has `Canvas#top`. The `<Canvas />` here cant exactly know its state information since we are using it outside Craft, so what should we do ?
   - Using @ackvf solution, the `<Canvas />` should just be a placeholder and just return perhaps a Fragment. 
   - More importantly, what should this Canvas render as its children ? Well, the JSON tells us that it should render `<p>I've been dropped here</p>`. Normally `<Canvas />` has access to the Craft's internal state (eg: the JSON), so we know what we should render here, but now without Craft, it has no access to the state, so we can only render what has been defined in the JSX initially (which is a `h3`). Unless, you somehow explicitly pass the portion of the JSON to the `<Canvas json={...} />`, which as you can imagine is pretty awkward.

The only way I can see this making sense is you don't actually render the User Component that is defined in the `type`. For example, if the type is `Hero`, then we tell our custom function to do something like this:
```jsx
const LiveHero = ({children}) => {
  return (
     <div>
        <h2>Hero</h2>
        {children}
     </div>
  )
}
if ( type == "Hero ) {
  const topCanvas = childCanvas["top"];
  const children = topCanvas.nodes;
  return <LiveHero>{children}</LiveHero>
}
```

In which case I think this might work, although it looks like a huge maintenance burden down the road. 

Anyways, we're getting a bit sidetracked here. To clarify, developers can ultimately do what they want with the JSON including using their own rendering methods. But my concern is that they will need to find a workaround to deal with the Canvas inside User Components such as what I've mentioned. 

With that said, I'm still unsure as to how this justifies having 2 JSON outputs.

- If we are sticking to the current JSON output, it's fine but it's just not as storage efficient. 
- The new one is more hierarchical and storage efficient, but deserialisation process will be more complex.

hdaggett:
Just following up with this conversation - I don't totally understand the complications with nested Canvas elements, but any update on a cleaner JSON input/output? I have several pages currently stored as SlateJS JSON that I'd like to convert to CraftJS JSON, but I'm hesitant to do this until CraftJS has better support for hierarchical JSON and serializing/deserializing like SlateJS (see https://docs.slatejs.org/concepts/09-serializing)

Hyperkid123:
@hdaggett so the issue here is, that using the tree output outside of craft JS, is "polluted" with craft specific component (canvas). It cannot be removed because it would not be possible to re-create the craft internal schema.

I am more against embedding the tree output as I can now actually understand what is going inside of craft JS.

The function itself is very simple to implement, it's just a basic recursion (10 lines of code) so you can do that easily in your codebase.

@prevwong @ackvf I think I will probably close this and probably just show an example of such function which would create the tree structure in docs. What do you think?

hdaggett:
@Hyperkid123 an example of a function to convert CraftJS's JSON into a tree structure would be great!

@prevwong I'll also need to implement a function that converts my SlateJS JSON to CraftJS JSON. I'm guessing I can just generate IDs for each node and assign the properties for `nodes`, `isCanvas`, `parent`, and `_childCanvas` as needed. Anything else I should be aware of?

prevwong:
@Hyperkid123 I agree. Let's close this for now, we'll revisit this in the future if necessary. An example in the docs would be great. If you're still interested, feel free to submit another PR for that.

@hdaggett Yeap, I think those are the main properties you need to take care of.

(p.s. Sorry for the slow response this past month, was busy wrapping up my final semester)

michaelsharpe:
Was there any progress on this? I don't see anything in the documentation about generating a craft agnostic JSON file from the designs.  I consider this a crucial use case to make this an editor usable on multiple projects, as many will not want to tie themselves to using craft as a rendering engine for layouts and may have extenuating use cases to build and maintain their own renderer.  

Awesome work by the way!  Very clean library.

nicosh:
> > I can't think of any viable usage of the JSON output outside of Craft
> 
> Our usecase is this
> 
> 1. devs create new blocks and prop editors
> 2. content administrators use craft.js UI to compose pages and store them as JSONs in database
> 3. customers visit a page which is rendered by our own render function from the stored JSON (without craft.js)
> 
> As mentioned, there can even be two JSON outputs, one simplified and efficient for the page rendering and the other with all metadata required for craft.js.

@ackvf I have a similar scenario, did you figured out how to render components without craft.js stuff or remove useless js?

Issue #21, opened 2020-01-18
actions.setNodeSelection in useEditor


ankri:
**Is your feature request related to a problem? Please describe.**
When trying out `craft.js` the first time I needed to select a node when clicking something. I expected it to work like 

```jsx
const { setNodeSelection } = useEditor();
const { parent } = useNode(node => ({ parent: node.data.parent }));
....
setNodeSelection(parent, true);
```

Thanks to the documentation I found the way to select nodes. 👍 

**Describe the solution you'd like**
If you are open to API changes I would like to open a PR (I have already implemented it in my fork) so that `useEditor` exposes `setNodeSelection(id: NodeId, isSelected: boolean)`. The exposed method would make it possible to programmatically select nodes without relying on `ref` and the `mousedown` handler.

PR #22, opened 2020-01-18
actions.selectNode in useEditor

ankri:
Hi,

this is my proposed solution for #21 

This exposes a new API for selecting nodes programmatically:

```
  const { actions } = useEditor();
  const {
    id,
    parent
  } = useNode(node => ({
    parent: node.data.parent
  }));

...

<Btn
  className="mr-2 cursor-pointer"
  onClick={() => {
      actions.setNodeSelection(parent, true);
  }}
>
  <ArrowUp />
</Btn>
```

I also added an example to the landing example:
image

to enable selecting the parent like in the [GrapeJS Demo](https://grapesjs.com/demo.html)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ out
 dist
 lib
 *-*.log
+.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,3 @@ out
 dist
 lib
 *-*.log
-.vscode
--- a/packages/core/src/hooks/useEditor.tsx
+++ b/packages/core/src/hooks/useEditor.tsx
@@ -4,7 +4,7 @@ import {
 } from "../editor/useInternalEditor";
 import { ConnectorElementWrapper } from "@craftjs/utils";
 import { useMemo } from "react";
-import { NodeId } from "interfaces";
+import { NodeId } from "../interfaces";
 
 type Overwrite<T, U> = Pick<T, Exclude<keyof T, keyof U>> & U;
 type Delete<T, U> = Pick<T, Exclude<keyof T, U>>;
--- a/packages/core/src/hooks/useEditor.tsx
+++ b/packages/core/src/hooks/useEditor.tsx
@@ -50,16 +50,8 @@ export function useEditor<S>(collect?: any): useEditor<S> {
   const actions = useMemo(() => {
     return {
       ...EditorActions,
-      selectNode: (nodeId: NodeId) => {
+      selectNode: (nodeId: NodeId | null) => {
         setNodeEvent("selected", nodeId);
-      },
-      clearNodeSelection: () => {
-        setNodeEvent("selected", null);
-      },
-      hoverNode: (nodeId: NodeId) => {
-        setNodeEvent("hovered", nodeId);
-      },
-      clearHoveredNode: () => {
         setNodeEvent("hovered", null);
       }
     };
--- a/packages/core/src/hooks/useEditor.tsx
+++ b/packages/core/src/hooks/useEditor.tsx
@@ -16,10 +16,7 @@ export type useEditor<S = null> = Overwrite<
       useInternalEditor<S>["actions"],
       "setNodeEvent" | "setDOM" | "replaceNodes" | "reset"
     > & {
-      selectNode: (nodeId: NodeId) => void;
-      clearNodeSelection: () => void;
-      hoverNode: (nodeId: NodeId) => void;
-      clearHoveredNode: () => void;
+      selectNode: (nodeId: NodeId | null) => void;
     };
     query: Delete<useInternalEditor<S>["query"], "deserialize">;
   }
--- a/packages/core/src/hooks/useEditor.tsx
+++ b/packages/core/src/hooks/useEditor.tsx
@@ -3,6 +3,8 @@ import {
   EditorCollector
 } from "../editor/useInternalEditor";
 import { ConnectorElementWrapper } from "@craftjs/utils";
+import { useMemo } from "react";
+import { NodeId } from "interfaces";
 
 type Overwrite<T, U> = Pick<T, Exclude<keyof T, keyof U>> & U;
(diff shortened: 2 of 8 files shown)


prevwong:
Thanks for the PR! 💯Just a few comments:

- I think `selectNode` might be a better name
- Since we already have `setNodeEvent` which is used internally, I think we should add this method to the [`useEditor`](https://github.com/prevwong/craft.js/blob/master/packages/core/src/hooks/useEditor.tsx) hook directly (see line 45, this is where the actions are attached to the hook. You can add the new method here). To keep things a bit clean looking, I prefer if the actions are written like in [useInternalNode](https://github.com/prevwong/craft.js/blob/master/packages/core/src/nodes/useInternalNode.ts) (line 32-39)
- For the arrow up example in the demo, could you place a condition to check if the selected Node has a parent (The Root Node does not have a parent) ? If there is no parent, don't display the button/icon.

p.s. I see that the files you've edited in the docs and examples are being formatted (by prettier?). I've not set up linting/formatting for the docs/examples, so for now please disable your formatter when editing those files.

ankri:
> I think `selectNode` might be a better name
I opted for `setNodeSelection` because you can select and deselect nodes with this API. With `selectNode` it would seem that one could only "select" nodes and not deselect them. But I don't mind using `selectNode` instead.

> To keep things a bit clean looking
Will do

> check if the selected Node has a parent 
What about `id !== ROOT_NODE` ?

> prettier
Yes. That's a problem. Could you please add a `.prettierrc` file to the root of the project? That will force my (and other's) prettiers to respect the original settings.

ankri:
I added your suggestions:
1. I renamed `setNodeSelection` to `selectNode: (nodeId: NodeId) => void` - but I added a second method `clearNodeSelection: () => void` so the user is able do deselect the currently selected node
2. I moved the actions to `useEditor` and used the pattern from `useInternalNode`
3. I added `hoverNode: (nodeId: NodeId) => void` and `clearHoveredNode: () => void` so that the user is able to also programmatically alter the hovered state. 
4. I used `clearHoveredNode` in the `RenderNode` and also added the `id !== ROOT_NODE` check
5. I deactivated prettier before adding the documentation

While trying out the `selectNode` function in the example it was confusing that the old node was still being hovered but the parent was now selected. That's why I also added  `hoverNode` and `clearHoveredNode` - but I do not really like the name `clearHoveredNode` - what do you think?

prevwong:
> I added `hoverNode: (nodeId: NodeId) => void`

I'm wondering what is the use case of needing to programatically alter a Node's hover state ? I thought `selectNode` made sense because when a user clicks on something, and something happens (eg: via Promise), then we can make the editor to select a different Node.

```jsx
<DeleteBtn
  onClick={() => {
    await fetch("deleteNode")
    // node deleted, select parent node
    selectNode(parentId)
  }
>
```

Since it's a matter of "which Node is being selected", I think it makes sense that we could programatically force it. But for hovering, I feel it's like it's the same as altering the `dragged` state, both feels "unnatural" to do that since we cannot guarantee that the Node is actually being hovered or dragged. 

It is technically possible, but I feel it has some semantics issue. (This was why I decided to hide `setNodeEvent` in the first place)

> I added a second method clearNodeSelection: () => void so the user is able do deselect the currently selected node

I prefer to keep it all in 1 function. Maybe `selectNode(false)` to deselect the currently selected Node ?

ankri:
Thanks again for your comments!

> I prefer to keep it all in 1 function

What do you think about `selectNode(nodeId: string | null)` and then calling `selectNode(null)` to clear the selection?

> `setHovered`

In this case I see `hover` as a way to highlight the component. In your layers example you hover the component in the layer list and it will be highlighted (hovered) in the Editor.

In my case I imagine some kind of [Omnibar](https://blueprintjs.com/docs/#select/omnibar) to search for components. You could enter the components name / prop / etc. into the omnibar and it would begin to filter the layers. Think of the movie search results in the Omnibar example to be the components. When a movie gets highlighted I want to highlight the component in the background. As soon as the component will be selected in the Omnibar I want to select the component in the Editor.

---

Not clearing the hovered component when selecting another component will result in two components being highlighted at the same time which can be confusing. [GrapeJs](https://grapesjs.com/demo.html) clears the currently hovered component, too when selecting the parent component. 

Maybe `selectNode` should automatically clear the hovered state. But I would still prefer not to hide `setHovered` because it's ability to highlight components.

prevwong:
> What do you think about selectNode(nodeId: string | null) and then calling selectNode(null) to clear the selection?

Ya, this is good too. The `setNodeEvent` accepts a null value which would deselect the Node, so you could just extend on that.

>  I imagine some kind of Omnibar to search for components.

Oh, this makes sense too. 

> Maybe selectNode should automatically clear the hovered state. 

Right, since you can't still be hovering on one Node when you've selected another Node. Maybe let's create a new PR for this, so we can let the changes in this PR be atomic.

ankri:
Back to the roots: this PR is now only about selecting nodes. Like you said. Let's keep it atomic.

Now it's just `selectNode(NodeId | null)`. Internally it resets the hovered state.

.gitignore
```diff
@@ -6,3 +6,4 @@ out
 dist
 lib
 *-*.log
+.vscode
```
prevwong: Remove this for now, I'll add this in a separate commit

packages/core/src/hooks/useEditor.tsx
```diff
@@ -3,6 +3,8 @@ import {
   EditorCollector
 } from "../editor/useInternalEditor";
 import { ConnectorElementWrapper } from "@craftjs/utils";
+import { useMemo } from "react";
+import { NodeId } from "interfaces";
```
prevwong: Change this to `.../interfaces`, otherwise the type definition for `selectNode` will be lost.

ankri:
> Just some minor changes needed, then we're all set 🎉😁

Done 👍

PR #24, opened 2020-01-19
Draft: History - undo/redo

ankri:
Hi,

last night I tried several ways to implement undo / redo functionality. I'd like to pick your brain on how to best tackle this problem in your opinion.

# Serializing / Deserializing the current state
The easiest way to achieve a undo / redo functionality is to serialize the current state in each action.

```ts
setProp(id: NodeId, cb: (props: any) => void) {
  invariant(state.nodes[id], ERROR_INVALID_NODEID);
  undoStack.push(query.serialize());
  cb(state.nodes[id].data.props);
},
``` 

and then expose a `undo` method and calling deserialize there

```
undo() {
  const undoState = undoStack.pop();
  if (undoState) {
    deserialize(undoState);
  }
},
```

Pros:
- Easy

Cons:
- The history stack can grow very large over time

# Using immer

Since you are using `immer` under the hood I tried to find a way to implement it with `PatchListener`s. But I ran into some problems:

### Adding a patchListener to useMethods
I tried to add a PatchListener to the source. But then all the events will be made `undoable` and I do not think hovered state needs to be undoable 😉.  After filtering out all `"events"` it was way less noisy but still too much. 

### Adding a patchListener in `actions.ts`

We could also add a patchListener directly in e.g. `setProp`

```
setProp(id: NodeId, cb: (props: any) => void) {
  invariant(state.nodes[id], ERROR_INVALID_NODEID);
  state.nodes = produce(state.nodes, (draftedNodes) => {
    cb(draftedNodes[id].data.props);
   }, (patches, inversePatches) => {
     undoStack.push(inversePatches);
  });
},
```

It works for setProp, but it does not work for `deleteNode`:
```
delete(id: NodeId) {
  invariant(id !== ROOT_NODE, "Cannot delete Root node");

  state.nodes = produce(
    state.nodes,
    draftedNodes => {
      const targetNode = draftedNodes[id];
      if (query.node(targetNode.id).isCanvas()) {
        invariant(
          !query.node(targetNode.id).isTopLevelCanvas(),
          "Cannot delete a Canvas that is not a direct child of another Canvas"
        );
        targetNode.data.nodes!.forEach(childId => {
          _("delete")(childId);
        });
      }

      const parentNode = draftedNodes[targetNode.data.parent],
        parentChildNodesId = parentNode.data.nodes!;

      if (parentNode && parentChildNodesId.indexOf(id) > -1) {
        parentChildNodesId.splice(parentChildNodesId.indexOf(id), 1);
      }
      updateEventsNode(state, id, true);
      delete draftedNodes[id];
    },
    (patches, inversePatches) => {
      undoStack.push(inversePatches);
    }
  );
},
```

I think it does not work because `parentChildNodesId.splice` is not part of the inversePatches. I think Immer does not recognize the changes made to the parentChildNodes.

Maybe you have some insights into immer and patches and have an idea?
--- 

I personally lean towards making use of serializing and deserializing. Maybe we could use [JSON-Patch](https://github.com/Starcounter-Jack/JSON-Patch) to calculate the patches needed on the diff between the undo state and the new state.

But I would prefer if history was exposed by `useEditor` directly and not part of the actions. So I tried to pass a history object to the Actions to be able to push to the undoStack.

```
export const Actions = (
  state: EditorState,
  query: QueryCallbacksFor<typeof QueryMethods>
  history: History
) => {
```

But `useMethods` does not allow a third parameter. And I do not know if it is or should be possible to alter `useMethods` to accept a third parameter.

I could also make the undo and redo stack globally available. But that feels code smelly.

# API of History

I propose this API for the `History`  - or maybe call it `UndoManager` 🤔

```
interface History {
  canUndo: () => boolean;
  canRedo: () => boolean;

  undo: () => void;
  redo: () => void;

  withoutUndo: (() => Function);
}
```

With `withoutUndo` you can skip adding items to the history.

```
// this will be undoable
actions.setProp(nodeId, (props) => { props.title = 'New title' });

// this will not be added to the history, thus not be able to be undoable
withUndo(() => {
  actions.delete(parentId);
});
```

--- 

If it's okay for you, I would gladly like to take over the implementation for the history.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ out
 dist
 lib
 *-*.log
+.vscode
--- a/packages/core/src/editor/actions.ts
+++ b/packages/core/src/editor/actions.ts
@@ -21,12 +21,55 @@ import invariant from "tiny-invariant";
 import { deserializeNode } from "../utils/deserializeNode";
 import { createElement } from "react";
 
+const undoStack: string[] = [];
+
 export const Actions = (
   state: EditorState,
   query: QueryCallbacksFor<typeof QueryMethods>
 ) => {
   const _ = <T extends keyof CallbacksFor<typeof Actions>>(name: T) =>
     Actions(state, query)[name];
+
+  const deserialize = (json: string) => {
+    const reducedNodes: Record<NodeId, SerializedNodeData> = JSON.parse(json);
+    const rehydratedNodes = Object.keys(reducedNodes).reduce(
+      (accum: Nodes, id) => {
+        const {
+          type: Comp,
+          props,
+          parent,
+          nodes,
+          _childCanvas,
+          isCanvas,
+          custom
+        } = deserializeNode(reducedNodes[id], state.options.resolver);
+
+        if (!Comp) return accum;
+
+        accum[id] = query.createNode(createElement(Comp, props), {
+          id,
+          data: {
+            ...(isCanvas && { isCanvas }),
+            parent,
+            ...(isCanvas && { nodes }),
+            ...(_childCanvas && { _childCanvas }),
+            custom
+          }
+        });
+        return accum;
+      },
+      {}
+    );
+
+    state.events = {
+      dragged: null,
+      selected: null,
+      hovered: null,
+      indicator: null
+    };
+    state.nodes = rehydratedNodes;
+  };
+
   return {
     setOptions(cb: (options: Partial<Options>) => void) {
       cb(state.options);
@@ -46,6 +89,12 @@ export const Actions = (
     },
     replaceNodes(nodes: Nodes) {
       state.nodes = nodes;
+    },
+    undo() {
+      const undoState = undoStack.pop();
+      if (undoState) {
+        deserialize(undoState);
+      }
     },
     reset() {
       state.nodes = {};
@@ -136,6 +185,8 @@ export const Actions = (
      * @param index
      */
     move(targetId: NodeId, newParentId: NodeId, index: number) {
+      undoStack.push(query.serialize());
+
       const targetNode = state.nodes[targetId],
         currentParentId = targetNode.data.parent!,
         newParent = state.nodes[newParentId],
@@ -163,6 +214,8 @@ export const Actions = (
      */
     delete(id: NodeId) {
       invariant(id !== ROOT_NODE, "Cannot delete Root node");
+      undoStack.push(query.serialize());
+
       const targetNode = state.nodes[id];
       if (query.node(targetNode.id).isCanvas()) {
         invariant(
@@ -190,6 +243,7 @@ export const Actions = (
      */
     setProp(id: NodeId, cb: (props: any) => void) {
       invariant(state.nodes[id], ERROR_INVALID_NODEID);
+      undoStack.push(query.serialize());
       cb(state.nodes[id].data.props);
     },
     /**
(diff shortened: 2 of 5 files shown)


prevwong:
I was planning on looking into this later on but since we're here now, I would be glad to have it solved even sooner 😂 I'm not sure if I completely understand the proposed History API. State changes are made with `actions`, but how would `History` even know that an action is being called (and what is the current state and next state) ?


Anyways, I've not actually tried patches yet so I'm unable to provide much insights, but from a very high-level overview, it seems like it would be a more efficient solution especially if we are interested in implementing something like a "realtime sync across devices" feature down the road.

From what've you described, it seems that the current issue is that we should not record the patches (or state changes) made by all actions (eg: ignore state changes made by `setNodeEvent`)

Maybe we can modify  `useMethods` (at [line 113](https://github.com/prevwong/craft.js/blob/b818b6f0626a209b24d92ec329ebd9890862eda9/packages/utils/src/useMethods.ts#L113)) we could implement something like this:
```jsx
const noUndo = ["setNodeEvent"];
let allPatches = [], allInversePatches = [];
return [
  (state: S, action: ActionUnion<R>) => {
    const query = queryMethods && createQuery(queryMethods, () => state);
    // return produce(
    //   state,
    //   (draft: S) => {
    //     if ( !noUndo.includes(action.type) ) {
                
    //     }

    //     methods(draft, query)[action.type](...action.payload);
    //   },
    // );

    const [nextState, patches, inversePatches] = produceWithPatches(
      state,
      (draft: S) => methods(draft, query)[action.type](...action.payload)
    );       

    if ( !noUndo.includes(action.type) ) {
      allPatches.push(patches);
      allInversePatches.push(inversePatches);
    }

    return nextState;
  },
  methods
];
```
From here, we could expose an `undo` and `redo` method to `useEditor` 🤔

ankri:
> I'm not sure if I completely understand the proposed History API

I modelled the proposed API after the [UndoManager](https://github.com/mobxjs/mobx-state-tree/tree/master/packages/mst-middlewares#undomanager) of mobx-state-tree. In my opinion it works great.

My bad.  I totally missed to describe the "private" part of the History API. This was supposed to be the public facing API.

The `History` would still expose something like

```
pushUndo(state: string)
pushRedo(state: string)
``` 

in case of using immer it would be
```
// in most cases the changes re grouped and meant to be made undoable all at once, hence the Array of Patches 
pushUndo(inversePatches: Patch[])
pushRedo(patches: Patch[])
```

When not using "immer" I would love to pass the "private" part of history to the actions. Before every action we would do `history.pushUndo(query.serialize())`. Sadly the serialized string gets large. But as soon as I have the time I will look into creating a JSON patch (just like immer would do).

 > From what've you described, it seems that the current issue is that we should not record the patches (or state changes) made by all actions (eg: ignore state changes made by setNodeEvent)

Sadly that seems to be only half of the problem. Yes it's true that too many patches will be collected.  And yes, you could filter out the events like you described. I tried that in an unpublished branch.

But it seems that it does not work for, e.g. `deleteNode`.

```
delete(id: NodeId) {
  invariant(id !== ROOT_NODE, "Cannot delete Root node");

  state.nodes = produce(
    state.nodes,
    draftedNodes => {
      const targetNode = draftedNodes[id];
      if (query.node(targetNode.id).isCanvas()) {
        invariant(
          !query.node(targetNode.id).isTopLevelCanvas(),
          "Cannot delete a Canvas that is not a direct child of another Canvas"
        );
        targetNode.data.nodes!.forEach(childId => {
          _("delete")(childId);
        });
      }

      const parentNode = draftedNodes[targetNode.data.parent],
        parentChildNodesId = parentNode.data.nodes!;
      
      // In my attempts this did not result in a patch
      if (parentNode && parentChildNodesId.indexOf(id) > -1) {
        parentChildNodesId.splice(parentChildNodesId.indexOf(id), 1);
      }
     
      updateEventsNode(state, id, true);
      // this did
      delete draftedNodes[id];
    },
    (patches, inversePatches) => {
      undoStack.push(inversePatches);
    }
  );
},
```

I don't know why right now (will look into that when I have the time) but the `splice` did not result in a patch. Which meant that the undo was incomplete.

---

In any case I would love to expose `history` from `useEditor`

```jsx
const UndoButon = () => {
  const { history } = useEditor();

  return (
    <button disabled={!history.canUndo()} onClick={() => { history.undo() }}>Undo</button>
  );
}
```

ankri:
Closing, since it has been implemented already

PR #29, opened 2020-01-27
[WIP] Package json

ackvf:
In the order of appearance:

There are some lerna specific keys, which probably don't work from `package.json`. They should be configured in `lerna.json` where they already actually are.

The `ignore` pattern should probably contain the `examples` folder instead?
Have you changed the folder structure in the past?

<s>Shouldn't the intermediary `packages` object be removed from `workspaces`?
I am not fluent with lerna, but when I searched resources to understand it, it seems that the syntax of `package.json` `workspaces` remains managed by `yarn`.
https://medium.com/@jsilvax/a-workflow-guide-for-lerna-with-yarn-workspaces-60f97481149d
https://legacy.yarnpkg.com/blog/2017/08/02/introducing-workspaces/</s>

I moved several blocks in package.json files so that they all follow same structure. It's then easier to read and maintain them.

Does it make sense to link sibling packages internally to save space and to make them use latest/local version of other packages? This is good for local development to be able to test the examples against a local forked version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,4 @@
 {
-  "npmClient": "yarn",
   "private": true,
   "author": "Prev Wong <prevwong@gmail.com>",
   "scripts": {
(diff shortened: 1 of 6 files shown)


ackvf:
I have just found about the new `workspaces` syntax here https://legacy.yarnpkg.com/blog/2018/02/15/nohoist/#how-to-use-it

Can I change the PR to Draft?

prevwong:
> There are some lerna specific keys, which probably don't work from package.json. They should be configured in lerna.json where they already actually are.

I don't see it, could you show me which keys specifically?

> The ignore pattern should probably contain the examples folder instead?

The command accepts a glob of package names to ignore. In this case, I specified `example-*` to ignore `example-basic` and `example-landing`.

> Does it make sense to link sibling packages internally to save space and to make them use latest/local version of other packages?

Yarn/Lerna already does this automatically for us.

Issue #26, opened 2020-01-20
Add targetNode to canDrag rule


ackvf:
I have a BoxListContainer and BoxListItem

The BoxListContainer is set to only accept BoxListItems
```
BoxListContainer.craft.rules.canMoveIn = incomingNode => incomingNode.data.type === BoxListItem
```

Now I need to also set that BoxListItem can be dragged only into BoxListContainer. How do I do that?
```
canDrag(currentNode: Node, helpers: NodeHelpers) => boolean
canMoveIn(incomingNode: Node, currentNode: Node, helpers: NodeHelpers) => boolean
canMoveout(outgoingNode: Node, currentNode: Node, helpers: NodeHelpers) => boolean
```

I am missing `targetNode` in `canDrag`, the other two have both nodes available.

this:
```
canDrag(currentNode: Node, targetNode: Node, helpers: NodeHelpers) => boolean

BoxListItem.craft.rules.canMoveIn = (_currentNode, targetNode) => targetNode.data.type === BoxListContainer
```

prevwong:
You can use the `NodeHelpers` for this:

```jsx
canDrag: (node, helpers) => {
   return helpers(node.parent).get().data.type == BoxListContainer
}
```

The `canDrag` method only deals with the Node that's trying to move, thus only that `Node` is exposed to the method. The other 2 rules deals with the incoming/outgoing child Node and the parent Canvas Node itself.

ackvf:
Nice, I noticed the helpers there and thought I could get the target node, thanks for pointing out `node.parent` to me. That will solve my issue.

Anyway, on the other hand, wouldn't it make sense to make the signature of all rules similar? After all, dragging is always done in relation, so I think it makes sense to put
1) the currently dragged element
2) the element (or perhaps only Canvas element, otherwise null) it hovers over


To defend this proposal, I think it will actually be quite common to do both:
1) set Canvas to accept certain Components
2) limit a Component to be droppable to certain Canvases

With this change, it will be much easier to write the logic, so instead of having to update, say, 3 Canvases to accept a new Component, we can simply implement the logic in the Component itself.

```
// now - 15 LOC

CardTop.craft = {
  rules: {
    canMoveIn: incomingNode => incomingNode.data.type == Text
  }
}

Carousel.craft = {
  rules: {
    canMoveIn: incomingNode => incomingNode.data.type == Text
  }
}

TextContainer.craft = {
  rules: {
[...]

PR #35, opened 2020-02-21
added rules.canDrop

ankri:
I also needed a `canDrop` rule like @ackvf and you discussed in #26 

This PR adds the `canDrop` rule:
```ts
{
  rules: {
    (targetNode: Node, currentNode, helpers: NodeHelpers) => boolean
  }
}
``` 

With this you can prevent the node that is being dragged (`currentNode`) to be dropped into the target node (`targetNode`).

I also added documentation and a simple example to the Basic example. In there I added two components: `<ColumnLayout />` and `<Column />`

image

`<Column />`s may only be added to `<ColumnLayout />`s and `<ColumnLayout />`s only accept `<Column />`s.

image
> An empty `<ColumnLayout />`

image
> A `<ColumnLayout />` with two `<Column />`s


ankri:
There is one problem with this approach: 
- Drag a component that has a `canDrop` rule (let's take Column from the example above)
- Drag this `Column` over a `ColumnLayout`
- While dragging `craft` will automatically add that node
- When dropping the `Column` into a `Card` which is not allowed by the `canDrop` rule the `Column` is still visible inside the `ColumnLayout`

I think that is a bug with `craft.js`. While dragging, nodes should not be automatically added. They should only be added after being dropped.

@prevwong Should I file a separate issue? If you have some pointers I can try to fix that bug, too :)

ackvf:
It was discussed previously, that the automounting of components is an unexpected behavior and that, for example the Grapejs library also doesn't automount components when they start dragging from the toolbox. I believe @prevwong expressed that he wants to resolve this by preventing the automount, so its safe to work on.

prevwong:
> I think that is a bug with craft.js

@ankri I've merged a fix for that in master, can you pull it to your local branch and try again?

ankri:
Finally merged with master. Finally ready to review.

Issue #28, opened 2020-01-21
Add draggedOver to state.events


ackvf:
/edit Title changed from
### Why is dragged item also hovered when dragging over another item?

This is not a bug per se, but still an unexpected and weird behaviour.
I am dragging an item from its former location to another location, my cursor is above another Container, but the original dragged item still remains `hovered` even though that doesn't really make sense. 
![image](https://user-images.githubusercontent.com/2313018/72822520-87c51580-3c72-11ea-9d4d-f6799037a0de.png)

I have discovered this weird behaviour while trying to work around the limitation in #26.
I desperately need to tell which element am I currently dragging over. I used this `custom` proxy hack to achieve that, maybe it's wrong and I am getting stale value there?

```
const BoxListItem: UserComponent<BoxListItemProps> = props => {
  const { id, connectors: { connect, drag } } = useNode();
  const { actions: { setCustom }, hovered } = useEditor(state => console.debug('hovered', state.events.hovered) || ({ hovered: state.nodes[state.events.hovered].data.type }) )

  setCustom(id, custom => custom.hovered = hovered)
 
  // return JSX
```
```
BoxListItem.craft = {
  rules: {
    canDrag: node => console.debug('node.data.custom.hovered === ', node.data.custom.hovered === BoxList, node.data.custom.hovered) || node.data.custom.hovered === BoxList
  }
}
```
```
// this is where the state keeps printing
export const PropertiesPanel = () => {
  const {  PropsPanel, name } = useEditor(state => {
    if (!state.events.selected) return;
    console.debug('state', state.events, state)
    // return collector
```

Even though I will work on the `canDrop` rule as mentioned in #26 to implement the functionality, I believe that the editor state should better reflect what events are happening.

prevwong:
The `hovered` state is triggered by the `mouseover` listener, so of course this is different from what you're trying to do, which requires Craft to know the element on `dragover`. 

I think this might be useful though. For example, to have a component to add a style to indicate that it's a drop target when you're dragging something over it. 

Since they're not exactly the same thing, `hovered` should be kept as is, but there should a new property (maybe simply `draggedOver` ? :laughing:)  to reference the `Node` that is being dragged over.

ackvf:
Thanks for clarification. After studying the Drag and Drop API for a bit, it makes sense.
`draggedOver` is a fitting name given the other states.

ankri:
I have started implementing a solution but ran into a problem. Can you guys give your opinion in #37 ?

PR #37, opened 2020-02-22
Added events.draggedOver

ankri:
This is a first attempt to solve #28. But is a draft only.

I started implementing a solution and added `events.draggedOver`. `useNode` now provides a selector for `events.draggedOver: boolean` and `useEditor` now provides a selector for `events.draggedOver: NodeId | null`. Currently this is implementation and example only. I have not added documentation yet because I ran into a problem:

Currently I implemented `draggedOver` like this:
https://github.com/prevwong/craft.js/blob/39350e5da1663a9b23a3060fc4f2e74786976011/packages/core/src/events/EventManager.tsx#L103-L107

It works great for my use case: Adding an outline to the Canvas that the Node will be dropped in.  The problem is that draggedOver will either be the id of the Component the Node is being dragged over if the node is a Canvas or it will be the id of the parent Canvas if the node is not a Canvas. 
It could break the developer's expectation because `draggedOver` will never be true for the component the node is being dragged over if the component is not a Canvas.

I couldn't come up with a good solution that set `draggedOver` to the dragged over `NodeId` while still being able to only add a CSS rule to the parent Canvas.

What do you think @prevwong ? Should we continue with `draggedOver` being either the `NodeId` or the `NodeId` of the parent canvas? Or go with `draggedOver` always being the `NodeId` of the node that is being dragged over?

---
Example Usage in the landing example [Container](https://github.com/prevwong/craft.js/blob/39350e5da1663a9b23a3060fc4f2e74786976011/packages/examples/landing/components/selectors/Container/index.tsx#L60-L87)
Example Usage in the landing example [OnlyButtons](https://github.com/prevwong/craft.js/blob/39350e5da1663a9b23a3060fc4f2e74786976011/packages/examples/landing/components/selectors/Custom1/index.tsx#L7-L39)

### DIFF
--- a/packages/core/src/editor/actions.ts
+++ b/packages/core/src/editor/actions.ts
@@ -53,7 +53,8 @@ export const Actions = (
         dragged: null,
         selected: null,
         hovered: null,
-        indicator: null
+        indicator: null,
+        draggedOver: null
       };
     },
     setDOM(id: NodeId, dom: HTMLElement) {
@@ -246,7 +247,8 @@ export const Actions = (
         dragged: null,
         selected: null,
         hovered: null,
-        indicator: null
+        indicator: null,
+        draggedOver: null
       };
       state.nodes = rehydratedNodes;
     }
--- a/packages/core/src/editor/store.tsx
+++ b/packages/core/src/editor/store.tsx
@@ -13,7 +13,8 @@ export const useEditorStore = (options): EditorStore => {
         selected: null,
         dragged: null,
         hovered: null,
-        indicator: null
+        indicator: null,
+        draggedOver: null
       },
       options
     },
--- a/packages/core/src/events/EventManager.tsx
+++ b/packages/core/src/events/EventManager.tsx
@@ -100,6 +100,12 @@ export const EventManager: React.FC = ({ children }) => {
             y: e.clientY
           });
 
+          if (query.node(id).isCanvas()) {
+            setNodeEvent("draggedOver", id);
+          } else {
+            setNodeEvent("draggedOver", query.node(id).get().data.parent);
+          }
+
           if (getPlaceholder) {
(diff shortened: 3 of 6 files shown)


prevwong:
Hey @ankri, what if we kept draggedOver to be the id of the Node, and let the developer to decide what they want to do with it? 

Let's use your example of adding an outline to the immediate Canvas of the Node that is being dragged  over:
```jsx

export const MyCanvas = () => {
    const {id} = useNode();
    const {isDraggedOver} = useEditor((state, query) => {
        let isDraggedOver = false;
        const { events: {draggedOver} } = state;
        if ( draggedOver ) {

           // Get all ancestor ids of the Node that is being dragged over
            const ancestors = query.node(draggedOver).ancestors();

            for ( let i = 0; i < ancestors.length; i++ ) {
                const isAncestorCanvas = query.isCanvas(ancestors[i]);
                if ( isAncestorCanvas ) {

                    // If the first Canvas ancestor is the current Canvas
                    if ( isAncestorCanvas == id ) {
                        isDraggedOver = true;
                    }
                    break;
                }
            }
        }
       return {
         isDraggedOver
       }
    });

    return (
        <div className={isDraggedOver ? 'outline' : ''}>
            ...
        </div>
    )
}
```

But yea, if this is a common use case then it might be a lot of code and not so convenient. Alternatively, maybe we can convert the example I wrote into a `NodeHelper`, so we can then easily do something like `query.node(id).hasChildNodeDraggedOver()`:

```jsx
export const MyCanvas = () => {
    const {id} = useNode();
    const {isDraggedOver} = useEditor((state, query) => ({
        isDraggedOver: query.node(id).hasChildNodeDraggedOver()
    }));

    return (
        <div className={isDraggedOver ? 'outline' : ''}>
            ...
        </div>
    )
}
```

ankri:
Closing in favor of #186

PR #47, opened 2020-03-20
Refactor Event Handling

### LINKED ISSUE #25: Some bug about event binding
1. The variable [`guarded`](https://github.com/prevwong/craft.js/blob/master/packages/utils/src/useHandlerGuard.ts#L15) in useHandlerGuard hooks is not wrapped by `useMemo` will create a new reference every time component render, and the next `useMemo` in [line 23](https://github.com/prevwong/craft.js/blob/master/packages/utils/src/useHandlerGuard.ts#L23)'s cache ability is failed, so it will recalc every time component render and return a new object.

2. After first time render finished(may be not first time, i means after craft finish all it's initialization process and ready to work), it bind 3 event listeners on every same element(i find it by chrome devtool), and after every time i do an operation on a element(such as move), craft will bind a another event listener on the element(some parent element also add an event listener).

I think the rebind bug's is caused by `useHandlerGuard` hook, but i can't find the real reason in it's long called chain(i haven't understood how craft work), can it would be resolved?
And no matter what, I believe use `useMemo` to wrap the `guarded` variable is necessary.

### LINKED ISSUE #46: Connectors will not work when the Editor is initially disabled
**Describe the bug**

When the `enabled` prop in `<Editor />` is initially set to false, all connectors fail to work even after the prop is set to true.

```jsx
const App = () => {
   const [enabled, setEnabled] = useState(false);

   return (
      <Editor enabled={enabled}
         ....
      </Editor>
   )
}

const Btn = () => {
    const { connectors } = useNode();
     // connectors wont work even after the Editor is enabled
    return <button ref={connectors.connect}>Click</button>
}
```

### Your environment

| craft.js           | v0.1.0-beta-3

prevwong:
This PR is intended to Improve on the overall structure of how events are handled in Craft. 

- More independence from React - removes the need for `useHandlerGuard` and `useConnectorHooks`. 
- Craft-specific DOM handlers are completely removed when the editor is disabled.
- Exposes `DerivedEventHandlers` which makes it easier to bootstrap new handlers/connectors.
- Fixes multiple event listeners bug. Closes #25
- Fixes bug that causes connectors to not work if the Editor is disabled initially. Closes #46


ankri:
Hi @prevwong will you include #35 and #37 in your refactor solution? Then I'll close them in favor of your solution here. 

Otherwise I'll try to find some time asap to merge #35 - on #37 I would need your feedback to continue.

If you're not in a hurry - would now be a good time to add some tests? Maybe some [cypress](https://www.cypress.io/) integration tests?

prevwong:
@ankri Let's keep those in a separate PR. Yeah, will be adding tests with Cypress but I will need some time for that.

PR #50, opened 2020-03-25
Improve project setup for craft

mresposito:
# New
* improve project setup for craft

# Test
* localhost


package.json
```diff
@@ -81,7 +81,7 @@
       "<rootDir>/packages/core/"
     ],
     "testMatch": [
-      "<rootDir>/packages/core/src/__tests__/**/?(*.)(spec|test).ts(x|)"
+      "<rootDir>/packages/core/src/tests/**/?(*.)(spec|test).ts(x|)"
```
prevwong: I would prefer to keep the `__tests__` folder as is, unless you have a reason that justifies the name change?

packages/core/src/render/Frame.tsx
```diff
@@ -17,7 +17,7 @@ export const Frame: React.FC<Frame> = ({ children, json }) => {
 
   const [render, setRender] = useState<React.ReactElement | null>(null);
 
-  const initial = useRef({
+  const initialState = useRef({
```
prevwong: Is there a reason why `initialState` would be better? In terms of readability, I believe it's pretty self-explanatory already, and the addition of "state" to its name might be misleading in regards to React as this is not a state variable.

mresposito: I don't find `initial` to be readable. It's like if I had a `begin` or `end` variable. WHat does it do? this is effectively the `initialState` of the `Frame` so I think it's overall a better name for this variable.

package.json
mresposito: `tests` is shorter to type rather than the awkward `__tests__`. and it makes your paths shorter.

There is also an argument that unit tests should be an essential part of a codebase, so they should follow a normal naming convention rather than an exceptional one.
Also note there are only two `__tests__` folders in this project, so migrating will be pretty easy.

prevwong: Personally, I don't find it awkward (and this is the default for Jest and most JS projects I've seen follow this convention as well). In contrary, I find having a 'test' folder that is alphabetically sorted with the other folders to be more distracting.

Overall, this seems to be a more of a subjective choice than an objective improvement.

mresposito: yaeh, makes sense!

ankri: I personally prefer to co-locate the tests (for heavily tested code bases)
```
  craft.js/packages/core/src/nodes/Canvas.tsx
  craft.js/packages/core/src/nodes/Canvas.spec.tsx
```

which would resonate with "unit tests should be an essential part of a codebase"

---

In less tested projects it is okay to have them in a separate folder. When keeping the tests in a separate folder I agree with @prevwong - I, too, think it's subjective and personally do not care whether it is `__tests__` or `tests`.

ankri:
I think it would also be a good time to introduce a `.prettierrc` and to add `prettier` to the [pre-commit hook](https://prettier.io/docs/en/precommit.html)

mresposito:
yeah, I'd be happy to add prettier to the `pre-commit`. Any preference on `write` vs `read` in prettier?

packages/core/src/render/Frame.tsx
prevwong: My only issue is that it's not actually a state value in `<Frame />`. What about renaming it to `initialProps` (or `initialPropsRef`) instead?

ankri:
> Any preference on `write` vs `read` in prettier?

I'm sorry. I do not know what you mean by that.

mresposito:
https://prettier.io/docs/en/cli.html yeah, that's quite confusing. I mean using `--check` or `--write`. Normally, it would be `--write`

ankri:
Ah thanks! I prefer `--write` since that's the beauty of prettier: Everyone can format the code the way they want it but it gets checked in with the same formatting.

prevwong:
@ankri @mresposito Sounds good. Let's maintain the existing eslint rules though (https://prettier.io/docs/en/integrating-with-linters.html)

packages/core/src/render/Frame.tsx
mresposito: sounds good

mresposito:
👍 I'll also upgrade to the latest version of prettier since they just released a `2.0.0`

package.json
```diff
@@ -81,7 +83,7 @@
       "<rootDir>/packages/core/"
     ],
     "testMatch": [
-      "<rootDir>/packages/core/src/__tests__/**/?(*.)(spec|test).ts(x|)"
+      "<rootDir>/packages/core/src/**/__tests__/?(*.)(spec|test).ts(x|)"
```
prevwong: @mresposito Let's maintain the `.spec` naming convention for test files. We can remove `|test` from the glob here for consistency.

mresposito: sounds good

prevwong:
Note: I've disabled Prettier for markdown files as it causes some issue with MDX (eg: when using React Elements in markdown, the extra line breaks are necessary) and I would like to have the flexibility in code examples to explain/highlight certain parts better.

PR #69, opened 2020-04-27
Merge changes from candulabs/craft.js

mresposito:
# New
* adds `onStateChange` callback to `Craft` editor
* Adds `Tree` data structure to inserting and deleting trees all at once
* Change onDropNode behaviour to parse the entire react node when we drop it into the dom
* improve unit testing setup, install enzyme as react testing library
* refactors EventHandler to be more readable, adds unit test
* refactor actions to have more atomic actions and adds unit tests


packages/core/src/editor/actions.ts
```diff
@@ -200,41 +263,42 @@ export const Actions = (
...
+    setProp(id: NodeId, cb: (props: any) => void) {
+      invariant(state.nodes[id], ERROR_INVALID_NODEID);
+      cb(state.nodes[id].data.props);
     },
-    deserialize(json: string) {
-      const reducedNodes: Record<NodeId, SerializedNodeData> = JSON.parse(json);
-      const rehydratedNodes = Object.keys(reducedNodes).reduce(
-        (accum: Nodes, id) => {
+
+    setState(dehydratedNodes: Record<NodeId, SerializedNodeData>) {
```
prevwong, line 277: I think we should fine a better name for this. `setState` gives the impression that we are changing specific state values in the editor, but we're actually replacing the entire state with the serialised data structure. Perhaps we can let this remain `deserialise` and just check if the parameter `dehydratedNodes` is a JSON string before-hand, in which case parse the JSON string and then continue on with the function.

packages/core/src/editor/query.tsx
```diff
@@ -152,14 +140,30 @@ export function QueryMethods(Editor: EditorState) {
...
     },
+
+    /**
+     * Get the current Editor options
+     */
+    getOptions(): Options {
+      return options;
+    },
+
+    getState(): Record<NodeId, SerializedNodeData> {
```
prevwong, line 151: Same as `deserialize` - since we're not actually get the actual `EditorState`, I feel this might cause confusion. Maybe, let it remain as `serialize`, with a parameter to specify if the return value should be in JSON? Although I feel we should omit the JSON capability in the future since it's pretty trivial, but for now it might help those who are expecting the value to be in JSON.

packages/core/src/editor/actions.ts
```diff
@@ -150,48 +171,90 @@ export const Actions = (
...
+      } else {
+        newParent.data.nodes = [targetId];
+      }
 
       state.nodes[targetId].data.parent = newParentId;
       state.nodes[targetId].data.index = index;
       currentParentNodes.splice(currentParentNodes.indexOf("marked"), 1);
     },
+
+    replaceEvents(events: EditorEvents) {
```
prevwong, line 185: Is there a use case for manually manipulating editor events (as oppose to using `setNodeEvent` and `setIndicator`) ?  Otherwise, I would combine this into `reset()` where we simply reset the editor events.

packages/core/src/index.tsx
```diff
@@ -1,6 +1,9 @@
...
-export * from "./hooks";
 export * from "./editor";
 export * from "./events";
+export * from "./hooks";
+export * from "./interfaces";
+export * from "./nodes";
+export * from "./render";
+
+// We expose this methods for candu's internal usage. Use at your own risksi
+export { useInternalEditor } from "./editor/useInternalEditor";
```
prevwong, line 9: Is there a reason for needing this (I'm assuming its to access a certain internal action)? Otherwise, let's remove it.

package.json
```diff
@@ -83,7 +85,7 @@
       "<rootDir>/packages/core/"
     ],
     "testMatch": [
-      "<rootDir>/packages/core/src/**/__tests__/?(*.)spec.ts(x|)"
+      "<rootDir>/packages/core/src/**/?(*.)test.ts(x|)"
```
prevwong, line 88: As with the previous PR, let's maintain `spec` for the naming convention, and `__tests__` for the containing folder :-)

packages/core/README.md
```diff
@@ -1,7 +1,10 @@
+## Important note!
+This is a fork of official [craft.js](https://github.com/prevwong/craft.js) repo. It's meant for internal @candulabs usage and might introduce some breaking changes. Use at your own risk, for stable version and support, please checkout main repo.
+
 
 <div align="center" style={{d}}>
 <h1>craft.js</h1>
-<a href="https://www.npmjs.com/package/@craftjs/core">
+<a href="https://www.npmjs.com/package/@craftjs/coreaft-core">
```
prevwong, line 7: Please discard all changes made to markdown files.

packages/core/src/editor/actions.ts
```diff
@@ -21,114 +22,134 @@ import invariant from "tiny-invariant";
...
-      const current = state.events[eventType];
-      if (current && id !== current) {
-        state.nodes[current].events[eventType] = false;
+    /**
+     * Add a new Node(s) to the editor.
+     *
+     * @param nodes
+     * @param parentId
+     */
+    add(nodes: Node[] | Node, parentId: NodeId) {
```
prevwong, line 61: I'm thinking of making this action simpler - as in to remove the need to support Node[], I initially used it only for `<Canvas />`, but still adding a single Node should suffice.

Perhaps we can remove this and rename `addNodeToIndex` as the new `add`

packages/core/src/editor/query.tsx
```diff
@@ -27,60 +29,47 @@ import {
...
+    createNode(reactElement: React.ReactElement | string, extras?: any) {
+      const node = transformJSXToNode(reactElement, extras);
+
       const name = resolveComponent(options.resolver, node.data.type);
       invariant(name !== null, ERRROR_NOT_IN_RESOLVER);
-      node.data.displayName = node.data.displayName
-        ? node.data.displayName
-        : name;
-
+      node.data.displayName = node.data.displayName || name;
```
prevwong, line 56: This will use the name from the resolver always, but we want to give priority to the name specified in the `.craft` property (which is stored in `.data.displayName`).

package.json
mresposito: Let's catch up about this, I'd love to maintain Candu's convention of `tests` and `test`. There are effectively no unit tests in Craft now, so let's maintain existing conventions.

packages/core/README.md
mresposito: ah my bad I thought I removed them.

packages/core/src/editor/query.tsx
mresposito: hmm I'm a bit confused.
I didn't change the logic here, I just removed a useless ternary for a more readable or statement. Isn't this statement doing what you want it to do?

packages/core/src/editor/query.tsx
mresposito: yeah I also feel the `getState` name is a bit confusing.

Most of the methods in `query` though have a getter notation, so I think we should also follow a `get____` something pattern like the rest of the getters.

Maybe `getSerializedNodes`? I'm not in love with it though, it's a mouthful. Also, if we call it`serialize` we'll never be able to deprecate the actual `serialize` method

packages/core/src/index.tsx
mresposito: ah good catch let me remove it.

packages/core/src/editor/query.tsx
prevwong: Oops sorry my bad, I got mixed up 😆

packages/core/src/events/EventHandlers.ts
```diff
@@ -1,9 +1,21 @@
...
+import { createShadow } from "./createShadow";
+import { NodeId, Indicator, Tree } from "../interfaces";
 import { Handlers, ConnectorsForHandlers } from "@craftjs/utils";
 import { debounce } from "debounce";
 import { EditorStore } from "../editor/store";
 
-type DraggedElement = NodeId | Node;
+type DraggedElement = NodeId | Tree;
+
+const event = ({
```
prevwong, line 9: Oh I like that this method. But could we have it follow a similar pattern like `addEventListener` instead of having to specify the object properties name/handler/capture? 

So like:
```jsx
const addEventListener = (name, handler, capture)  => {...}
```

Also, could you move this to `packages/utils/src/Handlers.ts` so that we can use this in the layers package as well.

packages/core/src/editor/query.tsx
prevwong: Hm, yeah but it is the quite descriptive haha. I don't mind `getSerializedNodes` for now with a deprecation notice on `serialize`

mresposito:
Abandoned in favor of #72

PR #70, opened 2020-05-06
fix: indicator width over the line

sprabowo:
Hi @prevwong 
please help to review this hotfix
Thanks

### DIFF
--- a/packages/layers/src/events/index.tsx
+++ b/packages/layers/src/events/index.tsx
@@ -49,7 +49,7 @@ export const EventManager: React.FC<any> = ({ children }) => {
         return {
           top,
           left: headingPos.left,
-          width: pos.width,
+          width: headingPos.width,
           height: 2,
           borderWidth: 0,
           background: color,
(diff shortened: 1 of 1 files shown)


packages/layers/src/events/index.tsx
```diff
@@ -49,7 +49,7 @@ export const EventManager: React.FC<any> = ({ children }) => {
         return {
           top,
           left: headingPos.left,
-          width: pos.width,
+          width: headingPos.width,
```
prevwong, line 52: Nice catch! But If we use the headingPos width, the indicator doesn't stretch fully till the rightmost edge. What about `width: pos.width - headingPos.left`?

sprabowo: I tried what you suggested yesterday and then I saw the line anomaly like didn't reach rightmost edge. agree with your suggestion. will update the code

Screen Shot 2020-05-07 at 16 21 52

PR #72, opened 2020-05-09
Improvements to PR-69

prevwong:
This PR is an extension to @mresposito's PR-69. 

Changes to original PR:
- Fixes `add` action
  - The original PR resets the parent Node's `data.nodes` whenever a new Node is added to it.
  - Handles `_childCanvas`. The original PR didn't handle this, so when users defines a `<Canvas />` inside a User Component, the state of that Canvas node is lost if the User Component is moved/rehydrated.
- Remove ability to process child Nodes of non-Canvas nodes in RenderNode
   - This will be addressed in the upcoming PR where we introduce the `<Element />` component that allows for non-Canvas Nodes to have child Nodes.
- Improve action/query method names
- Cherry picked `parseNodeFromReactNode` and `parseNodeFromSerialisedNode`
- Refactor and moved deprecation util to utils package


packages/utils/src/Handlers.ts
```diff
@@ -1,6 +1,10 @@
 import { wrapHookToRecognizeElement, Connector } from "./wrapConnectorHooks";
 
-export type HandlersMap<T extends string> = Record<T, Handler>;
+export const defineEventListener = (
```
mresposito, line 3: mmm it's a little wierd to put this util function in a separate package when it's only being used in a single fine and it's for a very specific purpose.

packages/core/src/events/EventHandlers.ts
```diff
@@ -32,43 +27,40 @@ export class EventHandlers extends Handlers<
       select: {
         init: () => () => this.store.actions.setNodeEvent("selected", null),
         events: [
-          event({
-            name: "mousedown",
-            handler: rapidDebounce((_, id: NodeId) =>
+          defineEventListener(
```
mresposito, line 30: mm is there a reason why you changed this? I find the LHS version slightly more readable (though it's more verbose), but it's not a strong preference since you can look up the function and find what each parameter does). I would move the `defineEventListener` function up to the top of this file though since you're not using it anywhere else anyways.

mresposito: btw, we should convert the underlying data structure to an object rather than an array. it's much more readable and extensible.

mresposito: let's leave that change for another PR though

packages/core/src/editor/store.tsx
```diff
@@ -4,17 +4,21 @@ import { QueryMethods } from "./query";
 
 export type EditorStore = SubscriberAndCallbacksFor<typeof Actions>;
 
+export const editorInitialState = {
```
mresposito, line 7: nice!

packages/core/src/editor/query.tsx
```diff
@@ -260,14 +288,35 @@ export function QueryMethods(state: EditorState) {
...
+     * @param node
+     */
+    getSerializedNode(node: Node): SerializedNode {
+      return serializeNode(node.data, options.resolver);
+    },
+
+    /**
+     * Returns all the `nodes` in a serialized format
+     */
+    getSerializedNodes(): SerializedNodes {
```
mresposito, line 307: `getSerializedNodesFromState` to differenciate from the one above? it's quite of a mouthful though

packages/core/src/editor/actions.ts
```diff
@@ -60,8 +57,11 @@ export const Actions = (
      */
     add(nodes: Node[] | Node, parentId: NodeId) {
       const parent = getParentAndValidate(parentId);
-      // reset the parent node ids
-      parent.data.nodes = [];
+
+      if (!parent.data.nodes) {
```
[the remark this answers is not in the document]
prevwong: I'm not resetting it but rather I'm creating the property because it doesn't exist yet.  This is because not all Nodes have child Nodes initially, so they don't necessarily have the `data.nodes` property specified. So what I did here was to initialised the property if it doesn't already exist. The same thing with `_childCanvas`.

packages/core/src/editor/actions.ts
```diff
@@ -1,46 +1,43 @@
...
 
 export const Actions = (
   state: EditorState,
   query: QueryCallbacksFor<typeof QueryMethods>
 ) => {
   /** Helper functions */
   const addNodeToParentAtIndex = (node: Node, parent: Node, index: number) => {
-    parent.data.nodes.splice(index, 0, node.id);
-    node.data.parent = parent.id;
+    if (parent && node.data.isCanvas && !parent.data.isCanvas) {
```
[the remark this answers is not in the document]
prevwong: This check is needed to know if the Node we are adding is supposed to be a childCanvas, in which case the parent is never a Canvas; whereas a child Node can only be added to a Canvas Node (Again this behaviour will be changed in the upcoming Element PR, but this is just temporarily so we don't break the current usage)

packages/core/src/editor/actions.ts
```diff
@@ -1,46 +1,43 @@
...
+      invariant(node.data.props.id, ERROR_ROOT_CANVAS_NO_ID);
+      if (!parent.data._childCanvas) parent.data._childCanvas = {};
```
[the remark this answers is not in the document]
prevwong: Hahha okay so let's say we have the following:

```jsx
export const Card = () => {
    const { connectors: {connect, drag} } = useNode()
    return (
        <div ref={dom => connect(drag(dom))}>
            <Canvas is={CardTop} id="top"> <-- this guy
                <h1>...</h1>
            </Canvas>
            <h5>...</h5>
        </div>
    )
}

export const CardTop = () => {
    const { connectors: {connect} } = useNode()
    return (
        <div ref={connect}>
            {children}
        </div>
    )
}
```

CardTop is not a child node of Card, but it's a `_childCanvas`. Child nodes are supplied to the component's `children` prop, and we know what to render beforehand since we have the child Nodes (and so we know the type and props to render).  This however is part of the component itself. So we don't know anything about CardTop until it is actually rendered, from there we will try to get it's Node by matching its `id` to it's parent `_childCanvas` property.

```
Card_Node: {
  type: Card,
  props: {...},
  _childCanvas: {
      "top" : "CardTop_Node" // When CardTop is rendered, we find its Node by looking it up here
  }
}
CardTop_Node: {
  type: CardTop,
  props: {...}
}
```
  
(p.s. The name of `_childCanvas` can be improved, but this is addressed in the upcoming Element PR)

packages/core/src/events/EventHandlers.ts
prevwong: I'm following a similar API as `addEventListener` which I think it's quite easily understood. I believe this should be kept in the utils package because the layers package makes use of Handlers as well

packages/core/src/editor/query.tsx
prevwong: Or perhaps let's rename `getSerializedNode` with the "parse" prefix similar to the other parse queries since we are not actually getting a state value but rather transforming an object into a different form. So maybe like `parseSerializedNodeFromNode`. 

It's mouthful but I like that it's descriptive. Maybe we can create a collection similar to NodeHelpers, eg: `query.parse.node.fromSerialised()` or something like that 😆

packages/core/src/editor/actions.ts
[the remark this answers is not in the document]
mresposito: thanks for the explanation! haha I do find this overall to be quite confusing, but let's tackle it in the next PR

packages/core/src/editor/actions.ts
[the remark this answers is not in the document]
mresposito: 👍

packages/core/src/editor/actions.ts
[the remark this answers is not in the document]
mresposito: 👍

packages/core/src/editor/query.tsx
mresposito: yeah I like the idea of the static node helpers, I think it will clean up some clutter. Buc yeah, let's keep it to another PR.
Ok let's change it to `parseSerializedNodeFromNode`, I think it makes sense.

packages/core/src/events/EventHandlers.ts
mresposito: ok, makes sense!

PR #74, opened 2020-05-25
feat: refactor API

### LINKED ISSUE #73: Every canvas is draggable
**Describe the bug**
It seems every `<Canvas />` element is draggable, regardless of whether or not it is an immediate child of another `<Canvas />`. While this may be hard to notice if the surface of every canvas is completely covered by other elements, like in a simple example, it causes lots of issues otherwise. 

**To Reproduce**

This seems to happen to any `<Canvas />` element not an immediate child of another `Canvas`, and becomes noticeable when that Canvas has a padding or a minimum size of some sort. This is noticeable with the root `Canvas` element immediately below the `Frame` element, as well as a `Canvas` element used to implement a simple styled container. A simplified version of the code from my project is as follows:

```js
export const Container = ({children}) => {
	const { connectors: {connect, drag} } = useNode();

	return (
		<div className="container" ref={ref => connect(drag(ref))}>
			<Canvas id="content">
				{children}
			</Canvas>
		</div>
	)
}
```

```css
.container {
	background: #0001;
	padding: 10px;
}

.container #content {
	padding: 10px;
}
```

In the above code, I expected half of the total 20px padding to act as the active region for placing elements inside the container, and the other half to act as the active region for placing elements outside the container. This part works, but the inside of the container becomes a draggable non-User Element, and dragging the container around only works as expected with the outer 10px of the padding. Additionally, dragging this inner 10px without activating the indicator and then letting go seems to crash the editor entirely.

**Expected behavior**
I expected Canvas nodes that are not immediate children of other Canvas elements to not be draggable, as mentioned in the documentation on the website.

### Your environment

| craft.js         | 0.1.0-beta.6
| React            | 16.13.1
| TypeScript       | 3.8.3
| Browser          | Firefox 75.0
| npm/Yarn         | 6.14.4
| Operating System | Solus 4.1

prevwong:
- `<Element />` deprecates the `<Canvas />` component
  - Non-Canvas nodes can now create and have child Nodes. But the existing concept of Canvas nodes defining droppable regions and having draggable elements still persists.
  - `<Element canvas>` defines a Canvas node and is the equivalent of `<Canvas />`
  - Custom properties can be passed via `<Element custom={{...}} />`
- `node.data._childCanvas` has been renamed to `node.data.linkedNodes`
- The `Tree` structure has been renamed to `NodeTree` for specificity 
- The `json` prop in `<Frame />` has been renamed to `data`. You can either pass the serialised nodes in its json string or its object form to `data`.
- The `onStateChange` prop in `<Editor />` has been renamed to `onNodesChange` for specificity.
- `query.createNode` has been deprecated
   - Replaced by `query.parseReactElement(...).toNodeTree()`
- Updated actions
  - `actions.add(nodes: Node[])` has been deprecated in favour of `actions.add(node: Node)`
  - `actions.addTree()` has been renamed to `actions.addNodeTree()`
  - `actions.parseNodeFromSerializedNode` has been renamed to `actions.parseSerializedNode().toNode()`
- Added new NodeHelper methods
  - `node(id).toSerializedNode()` gets the specified Node in it's serialised form
  - `node(id).toNodeTree()` returns the specified Node and it's child Nodes in a NodeTree 
- The DOM of non-canvas Nodes will not be made draggable (closes #73)
- Node IDs have been simplified
  - No longer prefixed with the type of Node (eg: `canvas-xxx` or `node-xxx`)
  - Root Node has been renamed to `ROOT` rather than `canvas-ROOT`. Existing serialised data that contains `canvas-ROOT` will still work, but it will be automatically renamed to `ROOT`

All deprecated methods/components/props will continue to work but will be removed in a future release.


packages/core/src/editor/NodeHelpers.ts
```diff
@@ -0,0 +1,189 @@
...
+  const node = state.nodes[id];
+
+  const ref = (id) => NodeHelpers(state, id);
+
+  const getNodeFromIdOrNode = (node: NodeId | Node) =>
+    typeof node === "string" ? state.nodes[node] : node;
+
+  return {
+    isCanvas() {
+      return !!node.data.isCanvas;
```
mresposito, line 31: btw at some ponint we should rething the `data` and also the `isCanvas` property.

packages/core/src/editor/NodeHelpers.ts
```diff
@@ -0,0 +1,189 @@
...
+} from "@craftjs/utils";
+import { serializeNode } from "../utils/serializeNode";
+import { mergeTrees } from "../utils/mergeTrees";
+
+export function NodeHelpers(state: EditorState, id: NodeId) {
+  invariant(typeof id == "string", ERROR_INVALID_NODE_ID);
+
+  const node = state.nodes[id];
+
+  const ref = (id) => NodeHelpers(state, id);
```
mresposito, line 24: could we call this `nodeHelpers` or something else? the name `ref` it's a bit confusing since you're calling the method with a different node id.

packages/core/src/editor/NodeHelpers.ts
```diff
@@ -0,0 +1,189 @@
...
+      deprecationWarning("query.node(id).isParentOfTopLevelCanvas", {
+        suggest: "query.node(id).isParentOfTopLevelNodes",
+      });
+      return this.isParentOfTopLevelNodes();
+    },
+    get() {
+      return node;
+    },
+    ancestors(deep = false) {
+      function recursive(id: NodeId, result: NodeId[] = [], depth: number = 0) {
```
mresposito, line 59: could we call this helper function in a bit of a more descriptive way?

mresposito: Same for the ones above.

packages/core/src/editor/NodeHelpers.ts
```diff
@@ -0,0 +1,189 @@
...
+        result.push(id);
+        const node = state.nodes[id];
+        if (deep || (!deep && depth === 0)) {
+          if (node.data.parent) {
+            result = recursive(node.data.parent, result, depth + 1);
```
mresposito, line 64: just return here, no?

packages/core/src/editor/NodeHelpers.ts
```diff
@@ -0,0 +1,189 @@
...
+                result.push(nodeId);
+                result = recursive(nodeId, result, depth + 1);
+              });
+          }
+
+          if (node.data.nodes) {
+            const childNodes = node.data.nodes; // ['t1c1', 't1c2']
+            childNodes.forEach((nodeId) => {
+              result.push(nodeId);
+              result = recursive(nodeId, result, depth + 1);
```
mresposito, line 88: just return...

packages/core/src/editor/NodeHelpers.ts
```diff
@@ -0,0 +1,189 @@
...
+        const targetNode = node;
+        invariant(!this.isTopLevelNode(), ERROR_MOVE_TOP_LEVEL_NODE);
+        invariant(
+          NodeHelpers(state, targetNode.data.parent).isCanvas(),
+          ERROR_MOVE_NONCANVAS_CHILD
+        );
+        invariant(targetNode.rules.canDrag(targetNode, ref), ERROR_CANNOT_DRAG);
+        return true;
+      } catch (err) {
+        if (onError) onError(err);
```
mresposito, line 110: expand bracket

packages/core/src/editor/NodeHelpers.ts
```diff
@@ -0,0 +1,189 @@
...
+              result.push(nodeId);
+              result = recursive(nodeId, result, depth + 1);
+            });
+          }
+        }
+        return result;
+      }
+      return recursive(id);
+    },
+    linkedNodes() {
```
mresposito, line 96: hmm what are linked nodes? I don't think I've seen this before

packages/core/src/editor/NodeHelpers.ts
```diff
@@ -0,0 +1,189 @@
...
+
+        invariant(currentParentNode.data.isCanvas, ERROR_MOVE_NONCANVAS_CHILD);
+
+        invariant(
+          currentParentNode ||
+            (!currentParentNode && !state.nodes[targetNode.id]),
+          ERROR_DUPLICATE_NODEID
+        );
+
+        const targetDeepNodes = NodeHelpers(state, targetNode.id).descendants();
```
mresposito, line 142: shouldn't you use ref here?

packages/core/src/editor/NodeHelpers.ts
```diff
@@ -0,0 +1,189 @@
...
+            targetNode,
+            currentParentNode,
+            ref
+          ),
+          ERROR_MOVE_OUTGOING_PARENT
+        );
+
+        return true;
+      } catch (err) {
+        if (onError) onError(err);
```
mresposito, line 160: expand pls

packages/core/src/editor/NodeHelpers.ts
```diff
@@ -0,0 +1,189 @@
...
+    toNodeTree() {
+      const childNodes = (node.data.nodes || []).map((childNodeId) => {
+        return NodeHelpers(state, childNodeId).toNodeTree();
+      });
+
+      return mergeTrees(node, childNodes);
+    },
+
+    /**
+            Deprecated NodeHelpers
```
mresposito, line 176: fix formatting

packages/core/src/editor/NodeHelpers.ts
```diff
@@ -0,0 +1,189 @@
...
+        **/
+
+    decendants(deep = false) {
+      deprecationWarning("query.node(id).decendants", {
+        suggest: "query.node(id).descendants 🙈",
```
mresposito, line 181: monkey is cute, but it's quite confusing :)

packages/core/src/editor/actions.ts
```diff
@@ -27,22 +28,61 @@ export const Actions = (
...
+  ) => {
+    const parent = getParentAndValidate(parentId);
+    // reset the parent node ids
+    if (!parent.data.nodes) parent.data.nodes = [];
+
+    if (parent.data.props.children) {
+      delete parent.data.props["children"];
     }
 
+    if (index != null) parent.data.nodes.splice(index, 0, node.id);
```
mresposito, line 44: let's expansd for readibily tpls?

packages/core/src/editor/actions.ts
```diff
@@ -27,22 +28,61 @@ export const Actions = (
...
-      parent.data.nodes.splice(index, 0, node.id);
-      node.data.parent = parent.id;
+  const addNodeToParentAtIndex = (
+    node: Node,
+    parentId: NodeId,
+    index?: number
+  ) => {
+    const parent = getParentAndValidate(parentId);
+    // reset the parent node ids
+    if (!parent.data.nodes) parent.data.nodes = [];
```
mresposito, line 38: here too

packages/core/src/editor/actions.ts
```diff
@@ -54,42 +94,43 @@ export const Actions = (
...
-      if (!parent.data.nodes) {
-        parent.data.nodes = [];
+      if (!parent.data.linkedNodes) {
+        parent.data.linkedNodes = {};
       }
 
-      if (parent.data.props.children) {
-        delete parent.data.props["children"];
-      }
+      parent.data.linkedNodes[id] = tree.rootNodeId;
```
mresposito, line 107: what what what
Can you update the description to mention why we need this?

packages/core/src/nodes/Element.tsx
```diff
@@ -0,0 +1,96 @@
...
+      id: node.id,
+      data: node.data,
+    },
+  }));
+
+  const [internalId, setInternalId] = useState<NodeId | null>(null);
+  const [initialised, setInitialised] = useState(false);
+
+  useEffectOnce(() => {
+    invariant(id != null, ERROR_ROOT_CANVAS_NO_ID);
```
mresposito, line 51: `!==`

packages/core/src/nodes/Element.tsx
```diff
@@ -0,0 +1,96 @@
...
+    if (inNodeContext) {
+      let internalId,
+        newProps = props;
+
+      const existingNode =
+        data.linkedNodes &&
+        data.linkedNodes[id] &&
+        query.node(data.linkedNodes[id]).get();
+
+      if (existingNode) {
```
mresposito, line 63: `existingNode && existingNode.data.type === props.is && typeof props.is !== "string"`

packages/core/src/render/Frame.tsx
```diff
@@ -22,29 +17,36 @@ export const Frame: React.FC<Frame> = ({ children, json, data }) => {
 
   const [render, setRender] = useState<React.ReactElement | null>(null);
 
+  if (!!json) {
```
mresposito, line 20: nice!

packages/core/src/tests/fixtures.ts
```diff
@@ -1,72 +1,59 @@
...
-  data: {
-    props: { childrenString: "Header 1" },
-    name: "Text",
-    displayName: "Text",
-    custom: {},
-  },
-  related: {},
-  events: { selected: false, dragged: false, hovered: false },
-  rules: {},
+export const createTestNode = (id, data, config = {}) => {
```
mresposito, line 5: oooh this is a useful one. Could you put this in utils maybe so we can also use it in other places?

packages/core/src/editor/NodeHelpers.ts
prevwong: So I remembered another reason why I kept the `type` prop in `data` - it's because when we're serialising, it's just a lot easier to to just get everything form `data`. About `isCanvas`, I'm thinking if we should rename it to just `canvas` to match `<Element />`'s canvas prop

packages/core/src/editor/actions.ts
prevwong: Take a look at the `<Element />` component. We get the NodeTree from the React Element, and create a new Node. We attach the new Node to the parent Node's `linkedNodes` property.

packages/core/src/editor/NodeHelpers.ts
prevwong: Please check the updated [docs](https://deploy-preview-74--craftjs.netlify.app/r/docs/concepts/user-components#defining-editable-elements. It's basically when you define Nodes (using the `<Element />`)  inside a User Component's render, they are different from child Nodes in that they are not passed to the component's render.

packages/docs/docs/concepts/user-components.md
```diff
@@ -212,89 +205,85 @@ const Toolbar = () => {
...
   )
 }
+```
 
-HeroHeader.craft = {
-  rules : {...}
-}
+> You must specify the `id` prop when defining new Nodes with `<Element />` inside a User Component
+
+We call these linked nodes since they are linked to another Node via an arbitary `id`. In this case, the `Text` node is linked to `Hero`'s node via its "title" id.
```
matdru, line 258: "title" id sounds a bit confusing, almost like title prop serves as an id -> shall we change this id to be "title_text" just to make that slightly more obvious

packages/utils/src/useMethods.ts
```diff
@@ -212,12 +212,9 @@ class Watcher<S> {
   }
 
   notify() {
-    // Give unsubscribing the priority. Any better way?
-    setTimeout(() => {
-      for (let i = 0; i < this.subscribers.length; i++) {
-        this.subscribers[i].collect();
-      }
-    });
+    for (let i = 0; i < this.subscribers.length; i++) {
```
matdru, line 215: should we convert this to `forEach` ?

packages/docs/docs/api/Editor.md
```diff
@@ -33,13 +34,13 @@ const App = () => {
       <Editor>
         <h2>My Page Editor</h2>
         <Frame> 
-          <Canvas is={Container}> // defines the Root Node
+          <Element is={Container}> // defines the Root Node
             <h2>Drag me around</h2>
             <MyComp text="You can drag me around too" />
-            <Canvas is="div" style={{background: "#333" }}>
+            <Element is="div" style={{background: "#333" }}>
```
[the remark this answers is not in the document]
matdru: I mean, the top one

packages/core/src/editor/query.tsx
```diff
@@ -318,5 +137,86 @@ export function QueryMethods(state: EditorState) {
     serialize(): string {
       return JSON.stringify(this.getSerializedNodes());
     },
+
+    parseReactElement: (reactElement: React.ReactElement) => ({
+      toNodeTree(
+        normalise?: (node: Node, jsx: React.ReactElement) => void
```
matdru, line 143: normalize*

Issue #82, opened 2020-06-16
onClick event not working in material-ui components


dadwic:
onClick event not working in user components but onMouseDown event it works!
e.g. Text component in [basic examples](https://craft.js.org/examples/basic/)

### Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Material-UI   |   4.10.2
| craft.js         | 0.1.0-beta.7
| React            | 16.8.6
| Chrome          | 83
| npm         | 6.14.4
| macOS | 10.15.4

@prevwong 
@ankri

prevwong:
So this is because the `connect`/`select` connector adds a `onClick` event handler to the DOM element and it prevents propagation, so this means any other `onClick` handler that you add will not be fired 😢 

We would definitely like to improve this behaviour. But in the meantime, the temporary workaround is to use a side effect that reacts to the `selected` event:

```jsx
const Button = () => {
  const { selected } = useNode(node => ({ selected: node.events.selected });

  useEffect(() => {
    if ( selected ) {
      console.log("Button clicked!")
   } 
  }, [selected]);
} 
```

dadwic:
Thanks,
In the previous version (v0.1.0-beta.6), this bug did not exist!

prevwong:
Yeah, previously for the `select` handler, we were using `onMouseDownCapture` with debounce, which made it possible for you to add your own event handlers without issues. 

But this had some caveats for us internally. For example, the event handler would be fired multiple times (so in the case of the `select` handler, we will be calling `setNodeEvent` multiple times, which was not ideal)

ymoussaba:
I was dealing with this issue after an update too. I use a popover to edit the components. But since most react libraries ([Antd](https://github.com/ant-design/ant-design/) in my case) are relying on **onClick**, nothing works.

Wouldn't there be a way to use another event than the onClick?

matdru:
I think the most sensible solution for time being will be to re-enable onClick propagation and introduce an internal flag system so that craft can handle selection events properly, @prevwong i can try and draft a PR for that

hdaggett:
> ```js-jsx
> const Button = () => {
>   const { selected } = useNode(node => ({ selected: node.events.selected });
> 
>   useEffect(() => {
>     if ( selected ) {
>       console.log("Button clicked!")
>    } 
>   }, [selected]);
> } 
> ```

This doesn't work for components that have multiple elements that may be clicked inside them :(

prevwong:
[...]

PR #85, opened 2020-06-20
Custom event blocking flag

matdru:
## New
- Added prettier config with no comma and single quote settings
- Event handler no longer stop propagation, using new flag to mark events as handled
- Adds helper script to package.json to ensure lerna links are working

## Fixes
- #82


prevwong:
@matdru  Do you mind opening a separate PR for prettier? :smile:  I try to keep PRs atomic whenever possible so it helps reverting them and facilitates discussions, even after the PR gets merged

matdru:
Yeah, sorry about that 😅

packages/utils/src/Handlers.ts
```diff
@@ -1,10 +1,23 @@
 import { wrapHookToRecognizeElement, Connector } from "./wrapConnectorHooks";
 
+// TODO(mat) any reason why why use array to store this information ?
```
matdru, line 3: @prevwong not sure if this structure is easy to read, any arguments why not use object instead of array?

prevwong: I just felt that in the class that would extend the Handlers class, typing it in an object form would be a bit less intuitive:

```
select: {
 events: [
    {
       name: "mousedown",
       handler: () => {...},
       capture: true
    }
 ]
}
```

Additionally, I wanted to have it resemble the native `addEventListener` API which an array was better able to achieve that `['mousedown', () => {...}, true]`

matdru: I see, well that makes sense, but now we lost this sync, since the third argument became an object so its no longer possible to use them as-is, maybe worth re-evaluating?

package.json
```diff
@@ -17,7 +17,10 @@
     "publish:beta": "lerna version prerelease --preid beta --force-publish --yes && release --pre",
     "test": "jest",
     "test:watch": "cross-env NODE_ENV=test jest --watchAll",
-    "lint": "eslint . --ext .js,.jsx,.ts,.tsx"
+    "lint": "eslint . --ext .js,.jsx,.ts,.tsx",
+    "scaffold": "npm run lerna:bootstrap && npm run lerna:link",
+    "lerna:link": "lerna link",
+    "lerna:bootstrap": "lerna bootstrap"
```
prevwong, line 23: Could you put this in the prettier PR instead?

packages/utils/src/Handlers.ts
prevwong: Yeah I agree, the third argument becoming an object takes away the magic, and we have the `defineEventListener` util which makes things easier for developers. 

But perhaps let's keep that in a separate PR? We need to update the `LayerHandlers` to use the `defineEventListener` util as well  (it's currently using the array form right now)

package.json
matdru: Ah, but its already closed 😅

prevwong: Yeh no worries, I can help with that. It should be separate from both PRs anyway

packages/core/src/events/EventHandlers.ts
```diff
@@ -24,32 +24,40 @@ export class EventHandlers extends Handlers<
       select: {
         init: () => () => this.store.actions.setNodeEvent('selected', null),
         events: [
-          defineEventListener('click', (e, id: NodeId) => {
-            e.stopPropagation();
-            this.store.actions.setNodeEvent('selected', id);
-          }),
+          defineEventListener(
+            'click',
```
prevwong, line 28: Let's switch back to `mousedown` here

packages/core/src/events/EventHandlers.ts
```diff
@@ -86,29 +95,32 @@ export class EventHandlers extends Handlers<
...
-                placement.parent.id,
-                placement.index + (placement.where === 'after' ? 1 : 0)
-              );
-            this.dropElement(onDropElement);
-          }),
+          defineEventListener(
+            'dragstart',
+            (e: DragEvent, id: NodeId) => {
+              // Ensure the Node is selected when it is being dragged
+              this.store.actions.setNodeEvent('selected', id);
```
prevwong, line 102: We can remove this if we are swithing back to `mousedown` for the select handler

PR #86, opened 2020-06-20
feat: improve NodeHelpers.descendants()

### LINKED ISSUE #84: node.toNodeTree() does not include linked nodes
I'm trying to create an editor for multiple pages using CraftJS. I have a single root node that has "page" nodes as children, and then I have components that can be put into pages.

I want to save each page separately, so I tried using `node.toNodeTree()` to extract the tree for a single page. However, it doesn't include linked nodes from named canvases. Do I need to recursively check for `node.data.linkedNodes` and add those to my tree? It seems like they should be included in `node.toNodeTree()`, but maybe I'm missing something.

Thanks!

prevwong:
This PR closes #84 

- Add option to include only childNodes or linkedNodes in `NodeHelpers.descendants()`
- Refactor `NodeHelpers.toNodeTree()` to use `NodeHelpers.descendants()`


packages/core/src/editor/NodeHelpers.ts
```diff
@@ -92,27 +91,29 @@ export function NodeHelpers(state: EditorState, id: NodeId) {
...
+            // Include child Nodes if any
+            if (childNodes) {
+              childNodes.forEach((nodeId) => {
+                result.push(nodeId);
+                result = appendChildNode(nodeId, result, depth + 1);
+              });
+            }
+          }
+
+          return result;
```
matdru, line 116: can we rename this result to 'descendantNodeIds' ?

prevwong: What about `descendants`? I think TS already makes it clear that it's `NodeId[]`

matdru: or that, sure 👍

PR #87, opened 2020-06-20
Add prettier config

matdru:
## New
- Prettier config
- switches to singleQuotes


.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
+  "trailingComma": "es5",
+  "tabWidth": 2,
+  "semi": false,
```
prevwong, line 4: I think you forgot to change this one 😛

matdru: Ah sorry, i missunderstood that we actually want to try semi-less aproach 😅

PR #118, opened 2020-08-20
Fix serializeNodes when props are undefined / null

tomgidden:
We're getting the error
```
index.js:362 TypeError: Cannot convert undefined or null to object
    at keys (<anonymous>)
    at Function.keys (es5-shim.js:1312)
    at serializeComp (index.js:1111)
    at index.js:1125
    at Array.reduce (<anonymous>)
    at serializeComp (index.js:1111)
    at serializeNode (index.js:1140)
    at Object.toSerializedNode (index.js:1288)
    at index.js:1387
    at Array.map (<anonymous>)
```
which appears to occur when props for an element (or a node within an element) is `null` or `undefined`, rather than `{}`.  I'm not sure exactly why this is happening at this point, but this seems to fix it.

### DIFF
--- a/packages/core/src/utils/serializeNode.tsx
+++ b/packages/core/src/utils/serializeNode.tsx
@@ -15,7 +15,7 @@ export const serializeComp = (
   resolver: Resolver
 ): ReducedComp => {
   let { type, isCanvas, props } = data;
-  props = Object.keys(props).reduce((result: Record<string, any>, key) => {
+  props = Object.keys(props || {}).reduce((result: Record<string, any>, key) => {
     const prop = props[key];
     if (!prop) {
       return result;


flavienrako:
Hi guys when can you fix it ?

flavienrako (approved):

matdru:
According to our TS interface and how craft was setup to work 'props' is not optional and should not be missing when serializing a node, and in such case i'm not sure if this PR is actually fixing the root issue here or just acting as a band-aid 🤔

prevwong:
Like what @matdru said, this error shouldn't have occurred in the first place, which suggests an underlying problem that we are not looking at. 

@tomgidden @flavienrako Does this issue this still happen in the latest version of Craft (0.1.0-beta.13)? If it does, could you share an example for us to reproduce the bug? Thanks!

PR #125, opened 2020-09-24
Fix handlers insertion when using an iframe

nicoladefranceschi:
This simple change allows using craftjs with an iframe that contains all the building nodes. 
It's important for me because I want to build a page that doesn't have the same style as the editor, so using an iframe allows to reset all the styles and start from scratch.

Some work needs to be done for the "packages/core/src/events/Events.tsx" file, because the "RenderIndicator" component should be rendered inside the iframe, not in the parent document.
I'm looking for a way to make the change as simple as possibile.

### DIFF
--- a/packages/core/src/events/Events.tsx
+++ b/packages/core/src/events/Events.tsx
@@ -29,6 +29,7 @@ export const Events: React.FC = ({ children }) => {
               : indicator.success,
             transition: '0.2s ease-in',
           },
+          parentDom: events.indicator.placement.parent.dom,
         })}
       {children}
     </EventHandlerContext.Provider>
--- a/packages/utils/src/Handlers.ts
+++ b/packages/utils/src/Handlers.ts
@@ -76,7 +76,7 @@ class WatchHandler {
     this.unsubscribe = store.subscribe(
       (state) => ({ enabled: state.options.enabled }),
       ({ enabled }) => {
-        if (!document.body.contains(el)) {
+        if (!el.ownerDocument.body.contains(el)) {
           this.remove();
           return this.unsubscribe();
         }
@@ -171,7 +171,7 @@ export abstract class Handlers<T extends string = null> {
       }
 
       const connector = (el, opts) => {
-        if (!el || !document.body.contains(el)) {
+        if (!el || !el.ownerDocument.body.contains(el)) {
           this.wm.delete(el);
           return;
         }
--- a/packages/utils/src/RenderIndicator.tsx
+++ b/packages/utils/src/RenderIndicator.tsx
@@ -1,7 +1,8 @@
 import React from 'react';
+import ReactDOM from 'react-dom';
 
-export const RenderIndicator: React.FC<any> = ({ style }) => {
-  return (
+export const RenderIndicator: React.FC<any> = ({ style, parentDom }) => {
+  const indicator = (
     <div
       style={{
         position: 'fixed',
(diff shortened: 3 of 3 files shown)


nicoladefranceschi:
I added another commit that fixes the RenderIndicator (it was rendered in the parent document, now it checks if we are inside an iframe, and if so, it create a React Portal inside the iframe)

andrewc910:
I agree craft.js should support iframes. I agree the editor shouldn't use the same styles as the webpage. However, I am a bit confused on how to use this.  I have cloned your branch, built & installed it. 

I am using the landing page example to test.

When I wrap the `iframe` around the `Editor` tag, nothing displays except the background, when i wrap around the `Viewport` tag or `Frame` tag, the toolbars display but not the editor. I also tried putting the iframe in the `Frame` tag but I don't think thats right. Any input would be appreciated.

khusseini:
Couldn't this be solved using styled components? I am building a framework around craft-js and use styled components for isolated styling: https://5f75849eea18900022958fa9-vdrllwtybd.chromatic.com/iframe.html?id=editor-viewport--usage&viewMode=story
Repository:
https://github.com/Akasha-Platform/craftjs-tools

Morphexe:
Any chance of this being merge ? Its actually a pretty useful feature.

prevwong (approved):
Look good! Thank you!

cederron:
This is not working in 0.2.x
Or am I doing something wrong ?

cederron:
> 

I was using the landing page example, I had to modify RenderNode createPortal to point to the iframe. Now it's working.
Thanks.

thirstycode:
Hello @nicoladefranceschi , Is there any example demonstrating how to use this in iFrame? I'm want to use it, but not sure how to implement the feature since there is mention about this in docs.

Andy-lm:
how to use the iframe，can provide a demo?

Andy-lm:
> Look good! Thank you!

how to use iframe in my project?can use provide me a demo?
 
is my code:
`<iframe
      onFocus={() => {
        console.log("======");
      }}
      ref={(dom: HTMLIFrameElement) => connect(dom)}
      style={{
        width,
        height: "calc(100vh - 36px)",
        backgroundColor: "green",
        // zoom
      }}
    >
    </iframe>`

LengYXin:
> 



> > Look good! Thank you!
> 
> how to use iframe in my project?can use provide me a demo?
> 
> is my code:
> 
> ```
> <iframe
>       ref={(dom: HTMLIFrameElement) => connect(dom)}
>     >
> </iframe> 
> ```

Didn't understand, could you be more detailed

piyushhanchate:
> > 
> 
> I was using the landing page example, I had to modify RenderNode createPortal to point to the iframe. Now it's working. Thanks.

Hey, could you post an example code of how you did this,please?

rajeshdavidbabu:
For those who are looking for iframe example. 

Check it out here https://codesandbox.io/p/sandbox/zen-roentgen-udqw3

rajeshdavidbabu:
For those who want to render craft.js contents into an iframe, here are my thoughts and learnings:

When you use native html <iframe> tags it would not work as expected as its an isolated context for both scripts and css. You could technically setup a postMessage communication layer between parent and child but that is too much work. So dont do this.

Ideally you should be using react-frame-component as advised by the Author, because what this library does is, it waits until your iframe is loaded and then renders your children inside a React portal . This React portal allows the contents rendered inside your iframe content to be part of your parent's React-tree.

Even if the iframe contents are part of React context, your external resources cannot be. So if your parent has an external .css file it would not work inside the react-frame-component's iframe but you may have to inject it seperately on the head of the iframe.

Here is an extended example of the Author's old example where you can see the contents rendered inside an iframe using react-frame-component but not styles isolated.

https://codesandbox.io/p/devbox/9tsc5r

PR #128, opened 2020-10-02
fix #49 Deserializing a Canvas with empty children raises an unhandled exception and fails

### LINKED ISSUE #49: Deserializing a Canvas with empty children raises an unhandled exception and fails
Hi,
First of all, I'd like to point out that this package is amazing.

We've been making use of it and encountered an issue when trying to serialize/deserialize our nodes.

A sample of the serialized output is:
```
{"canvas-ROOT":{"type":"div","isCanvas":true,"props":{"className":"page null"},"displayName":"Canvas","custom":{},"nodes":["canvas-jdl3oRkem"]},"canvas-jdl3oRkem":{"type":{"resolvedName":"Container"},"isCanvas":true,"props":{"children":null,"padding":20,"alignment":"column"},"displayName":"Canvas","custom":{},"index":0,"parent":"canvas-ROOT","nodes":["node-BbFEKEMN7"]},"node-BbFEKEMN7":{"type":{"resolvedName":"TextField"},"props":{"text":"Talks","fontSize":20},"displayName":"TextField","custom":{},"index":0,"parent":"canvas-jdl3oRkem"}}
```
When trying to deserialize it, we receive the following error:
```
Uncaught TypeError: Cannot read property 'resolvedName' of null
```

I have trace the issue to the following line:
https://github.com/prevwong/craft.js/blob/426e4962030e4443e53f004ce30da291b51992c0/packages/core/src/utils/deserializeNode.tsx#L37

What I believe that happens, is that as we can see, the node `canvas-jdl3oRkem` has a null children prop, null is truthy for `typeof null === "object"` however, it does not allow accessing it's attributes.

A possible solution would be to check if the prop value is null before trying to access it.

I'm also trying to understand why the null children prop is there to begin with.

What do you think?
@prevwong

vulcanoidlogic:
Set result[key] to null for  null / undefined prop


prevwong (approved):
Thanks!

Issue #110, opened 2020-08-02
SerializedNodes - Props - Missing Default Props value if it is empty(string) or false(boolean)


mkharibalaji:
**Describe the bug**
On executing query.getSerializedNodes() defaultProps values are missing if it is a empty String or when the boolean value is false.

**To Reproduce**
 1.Setup the basic example of Craftjs
 2.Add any UserComponent like Input/Select etc., with a defaultProps containing a empty String or a boolean value like this,
{required:false,helperText:'',label:'Test'}
3.onClick of the button run query.getSerializedNodes() to get the nodes of the above created user components and you will find the props with value of just {label:'Test'} and requried,helperText props are missing.

**Expected behavior**
DefaultProps should be available in the query.getSerializedNodes() even if the value is empty or false.

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| craft.js         | v0.1.0-beta.10
| React            | 16.13.1
| TypeScript       | 3.7.2
| Browser          | Firefox
| npm/Yarn         | Yarn
| Operating System | MacOS

PR #131, opened 2020-10-30
fix: serializeNode was naively checking if values were falsey

dbousamra:
This fixes https://github.com/prevwong/craft.js/issues/110

I haven't added any tests, as I noticed there were already none for this function. I can add some if you want.

### DIFF
--- a/packages/core/src/utils/serializeNode.tsx
+++ b/packages/core/src/utils/serializeNode.tsx
@@ -17,7 +17,8 @@ export const serializeComp = (
   let { type, isCanvas, props } = data;
   props = Object.keys(props).reduce((result: Record<string, any>, key) => {
     const prop = props[key];
-    if (!prop) {
+
+    if (prop == null) {
       return result;
     }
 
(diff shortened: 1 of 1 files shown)


packages/core/src/utils/serializeNode.tsx
```diff
@@ -17,7 +17,8 @@ export const serializeComp = (
   let { type, isCanvas, props } = data;
   props = Object.keys(props).reduce((result: Record<string, any>, key) => {
     const prop = props[key];
-    if (!prop) {
+
+    if (prop == null) {
```
prevwong, line 21: Good catch! 💯 Can you use strict equality `===` instead?

prevwong (changes requested):

packages/core/src/utils/serializeNode.tsx
dbousamra: Sure. Is triple equals the right way here? == will check against undefined, not just null.

prevwong: I mean something a bit more verbose like `prop === undefined || prop === null`. I think this makes the code a little bit more readable and potentially less bug prone.

prevwong (approved):
Thanks!

PR #133, opened 2020-11-03
Throw error when resolver is not an object +  Better error messages when deserializing unknown nodes 

ankri:
Yay I'm back. With a smaller change this time.

This should help to prevent some common beginner mistakes.

### Display an error when `resolver` on `<Editor />` is not an object
image

### Display an error which component cannot be resolved
image

### DIFF
--- a/packages/core/src/utils/deserializeNode.tsx
+++ b/packages/core/src/utils/deserializeNode.tsx
@@ -37,6 +37,7 @@ export const deserializeComp = (
 
   props = Object.keys(props).reduce((result: Record<string, any>, key) => {
     const prop = props[key];
+
     if (typeof prop === 'object' && prop.resolvedName) {
       result[key] = deserializeComp(prop, resolver);
     } else if (key === 'children' && Array.isArray(prop)) {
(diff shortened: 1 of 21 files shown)


ankri:
I replaced RTL with enzyme.  👍

packages/core/src/utils/deserializeNode.tsx
```diff
@@ -71,6 +74,15 @@ export const deserializeNode = (
...
   const { type: Comp, props: Props, ...nodeData } = data;
 
+  invariant(
+    (Comp !== undefined && typeof Comp === 'string' && Comp.length > 0) ||
+      (Comp as { resolvedName?: string }).resolvedName !== undefined,
+    ERROR_DESERIALIZE_COMPONENT_NOT_IN_RESOLVER.replace(
+      '%displayName%',
+      data.displayName
+    ).replace('%availableComponents%', Object.keys(resolver).join(', '))
+  );
```
prevwong, line 84: Can we break this line up so it is more readable? 

Ie:
```suggestion
   const isCompHtmlElement = Comp !== undefined && typeof Comp === 'string';
   const isCompUserComponent = Comp !== undefined && (Comp as { resolvedName?: string }).resolvedName !== undefined;

   invariant(isCompHtmlElement || isCompUserComponent,  '....')
```

prevwong (changes requested):
Welcome back! Looks good, just 1 tiny suggestion!

packages/core/src/utils/deserializeNode.tsx
ankri: good idea, done

prevwong (approved):

PR #148, opened 2020-12-04
feat: add Slate integration

prevwong:
This PR adds the ability to create RTE User Components via Slate.js


packages/slate/src/extend/slate/SlateNode.tsx
```diff
@@ -0,0 +1,57 @@
...
+  return (
+    <Slate editor={editor} value={value} onChange={setValue}>
+      <SlateNode.Provider value={{ id, enabled, actions }}>
+        {children}
+      </SlateNode.Provider>
+    </Slate>
+  );
+};
+
+export const useSlateNode = () => {
```
matdru, line 54: I'd maybe name this useSlateNodeContext for clarity

matdru (approved):

packages/slate/src/extend/slate/SlateNode.tsx
prevwong: I kinda prefer `useSlateNode` because:
- To resemble our `useNode` hook
- We may also want to extend this hook to contain additional functionality that's not directly part of the context value

PR #151, opened 2020-12-13
feat: add grid to styleInFlow

timc1:
Currently, `grid` user components have a different drag/drop interaction than `flex` – this PR adds `grid` to the list of display styles that enable `styleInFlow`.

### DIFF
--- a/packages/utils/src/getDOMInfo.ts
+++ b/packages/utils/src/getDOMInfo.ts
@@ -80,6 +80,7 @@ export const styleInFlow = (el: HTMLElement, parent: HTMLElement) => {
     case 'list-item':
     case 'table':
     case 'flex':
+    case 'grid':
       return true;
   }
   return;


prevwong (approved):
Thanks Tim!

timc1:
@prevwong any chance we can get picked into the `next` branch?

prevwong:
@timc1 As in to use it in your project? If so, you could use use the `0.2.0.alpha-x` releases on NPM. 
Unfortunately, we don't have documentation for these releases yet, but I would be happy to walk you through it, if you like (it's mostly related to the `state.events` being a `Set<NodeId>` rather than `NodeId` for multiselect).

timc1:
@prevwong Yep, we're working off the latest `0.2.0.alpha-x` releases, and would love for this fix to make it in: https://github.com/prevwong/craft.js/blob/next/packages/utils/src/getDOMInfo.ts#L82

matdru:
@timc1 version 0.2.x get automatically deployed from this branch -> https://github.com/prevwong/craft.js/tree/next , if you want to open PR with this change pointed there, it will make it in next release.

prevwong:
All good, I've taken care of this one. This should now be included in `0.2.0-alpha.13 `

PR #152, opened 2020-12-15
Improve drag previews

fredjens:
Adds possibility to set a custom image for drag preview.

Syntax:
```
connectors.create(
  ref,
  <Button text="Click me" size="small" />,
  'http://placekitten.com/200/200'  <- optional preview image
)
```

Inspired by react-dnd.

* Basic example updated
* Docs updated

### DIFF
--- a/packages/core/src/utils/Handlers.ts
+++ b/packages/core/src/utils/Handlers.ts
@@ -11,7 +11,7 @@ export type CraftDOMEvent<T extends Event> = T & {
 
 export type CraftEventListener = [
   string,
-  (e: CraftDOMEvent<Event>, opts: any) => void,
+  (e: CraftDOMEvent<Event>, opts: any, previewImage?: string) => void,
   boolean
 ];
 
@@ -26,12 +26,13 @@ export type Handler = {
    * The DOM manipulations to perform on the attached DOM element
    * @returns function that reverts the DOM manipulations performed
    */
-  init: (el: HTMLElement, opts: any) => any;
+  init: (el: HTMLElement, opts: any, previewImage?: string) => any;
 
   /**
    * List of Event Listeners to add to the attached DOM element
    */
   events: readonly CraftEventListener[];
+  previewImageUrl?: string;
 };
 
 export type ConnectorsForHandlers<T extends Handlers> = ReturnType<
@@ -64,6 +65,7 @@ const isEventBlockedByDescendant = (e, eventName, el) => {
 class WatchHandler {
   el: HTMLElement;
   opts: any;
+  previewImage?: string;
 
   private handler: Handler;
   private unsubscribe: () => void;
@@ -74,6 +76,7 @@ class WatchHandler {
     this.el = el;
     this.opts = opts;
     this.handler = handler;
+    this.previewImage = undefined;
 
     this.unsubscribe = store.subscribe(
       (state) => ({ enabled: state.options.enabled }),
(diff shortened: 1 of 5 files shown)


packages/examples/basic/components/Toolbox.js
```diff
@@ -66,6 +66,20 @@ export const Toolbox = () => {
             Card
           </MaterialButton>
         </Grid>
+        <Grid container direction="column" item>
+          <MaterialButton
+            ref={(ref) =>
+              connectors.create(
+                ref,
+                <Button text="Click me" size="small" />,
+                'http://placekitten.com/200/200'
```
ankri, line 75:
```suggestion
                'https://placekitten.com/200/200'
```

Otherwise the deployed Basic Example won't work on netlify:
image

packages/core/src/utils/Handlers.ts
```diff
@@ -26,12 +26,13 @@ export type Handler = {
...
    * @returns function that reverts the DOM manipulations performed
    */
-  init: (el: HTMLElement, opts: any) => any;
+  init: (el: HTMLElement, opts: any, previewImage?: string) => any;
 
   /**
    * List of Event Listeners to add to the attached DOM element
    */
   events: readonly CraftEventListener[];
+  previewImageUrl?: string;
```
prevwong, line 35: Thanks @fredjens for this PR, I really like this one! 

I'm just not too sure about adding `previewImage` directly into the internal API of the event handlers. However, this is the fault of the current state of the internal handlers, which definitely requires some refactoring.

My original idea for the API of the connectors was something like so:

```tsx
connectors.connectorName<T>(dom: HTMLElement, opts: T)
```

Where `opts` could be an object that we can pass to the connector. But it just so happens that almost every connectors' `opts` happened to just be a single value which is `nodeId` most of the time (ie: `connectors.drag(dom, nodeId)`.

Anyways, I'm thinking we could do either one of the following to fix this situation:

### Solution 1: Add a 3rd parameter for additional options

In this case, we basically turn the 2nd parameter to contain the required value by the connector (ie: the `nodeId`) and we add a 3rd parameter for additional options
 
```tsx
connectors.connectorName<R, T>(dom: HTMLElement, nodeId: R, moreOptions?: Partial<T>)

// Hence:
connector.drag(dom, nodeId) // or
connector.drag(dom, nodeId, { previewImage })

connector.create(dom, <UserComponent />) // or
connector.create(dom, <UserComponent />, { previewImage })
```

### Solution 2: Create an alias for the 2nd parameter

In this case, the connectors' 2nd parameter can either be the single required value (ie: `nodeId`) or an object to specify required value and additional options.

```tsx
connectors.connectorName<R>(dom: HTMLELement, nodeId: R);
connectors.connectorName<R, T>(dom: HTMLELement, { nodeId, previewImage }: {nodeId: R} & Partial<T>); 

// Hence:
connector.drag(dom, nodeId) // or
connector.drag(dom, {nodeId, previewImage })

connector.create(dom, <UserComponent />) // or
connector.create(dom,  { element: <UserComponent />, previewImage })
```

Personally, I prefer the first solution because it's probably easier to achieve; and the only benefit of the 2nd solution is that we keep parameter count to just 2, but at the expense of added complexity. What do you think?

prevwong (changes requested):

packages/core/src/utils/Handlers.ts
ankri: I prefer solution 1, too.

I don't think the added complexity is worth it. And I would argue, that from an API consumer's  perspective (esp. beginners)

```ts
// this is more readable
connectors.connectorName<R, T>(dom: HTMLElement, nodeId: R, moreOptions?: Partial<T>)

// than the "overload" version
connectors.connectorName<R>(dom: HTMLELement, nodeId: R);
connectors.connectorName<R, T>(dom: HTMLELement, { nodeId, previewImage }: {nodeId: R} & Partial<T>); 
```

fredjens: Hi there! Thanks for the feedback! Really sorry, but I do not think I have time to finish this in the nearest future... So if anyone else want to finish this, feel free!

PR #154, opened 2020-12-28
added touch events 

Nikshitha-ui:
made changes to support touch events using 'touch-dnd-custom-events ' depandency


packages/core/src/events/DefaultEventHandlers.ts
```diff
@@ -39,6 +40,7 @@ export class DefaultEventHandlers extends CoreEventHandlers {
   }
 
   handlers() {
+    setupTouchDNDCustomEvents();
```
prevwong, line 43: Hi @Nikshitha-ui, thanks again for this PR! 

I don't entirely understand why do we need this? It seems to me that we are already manually setting up the touch events.

Nikshitha-ui: This actually converts the touch events into drag events. That is onTouchStart mouse event is converted to onDragStart so that the element could be dragged similarly every touch events like onTouchMove is converted to onDragMove and so on

prevwong: I see. Would it be too much work to just handle the touch events without the need for the external library?

masonbrothers: From what I can tell, this was added because `DragEvent` inherits from `MouseEvent`. https://developer.mozilla.org/en-US/docs/Web/API/DragEvent

Since `DragEvent` does not inherit from `TouchEvent` (see https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent) it is difficult to use `DragEvent` on a touchscreen without some kind of polyfill.

PR #160, opened 2021-01-13
fix: typo

cbbfcd:


prevwong:
Hi @cbbfcd , thank you the PR! However, this is not a typo, the `create` connector is indeed intentionally used in this particular example.

More info on connectors here: https://craft.js.org/r/docs/api/useEditor

cbbfcd:
Thanks for reply and explanation

now: `const { connectors: { drag }} = useEditor();` 

it should be: `const { connectors: { create }} = useEditor();` 

I mean there should be a `create` method structured here. 

😂

PR #173, opened 2021-02-02
feat: Remove redundant code

cbbfcd:


packages/utils/src/getDOMInfo.ts
```diff
@@ -29,26 +37,22 @@ export const getDOMInfo = (dom: HTMLElement) => {
...
-    x: Math.round(x),
-    y: Math.round(y),
-    top: Math.round(top),
-    left: Math.round(left),
-    bottom: Math.round(bottom),
-    right: Math.round(right),
-    width: Math.round(width),
-    height: Math.round(height),
-    outerWidth: Math.round(width + margin.left + margin.right),
-    outerHeight: Math.round(height + margin.top + margin.bottom),
```
prevwong, line 41: Looks good, but can we keep `Math.round`? I don't think we need a separate `round` function just yet (even if we do, it shouldn't be exported from this `getDOMInfo` file as that would have a different responsibility)

cbbfcd: done.

prevwong (approved):
Thanks! 🥳

PR #174, opened 2021-02-03
fix: css grid drag and drop jank

timc1:
Follow up from https://github.com/prevwong/craft.js/pull/151.

This PR fixes the CSS grid drag/drop jank:

https://user-images.githubusercontent.com/12195101/106777448-ff15df00-6612-11eb-998c-6cf5a4eaa000.mp4

### DIFF
--- a/packages/utils/src/getDOMInfo.ts
+++ b/packages/utils/src/getDOMInfo.ts
@@ -55,6 +55,9 @@ export const styleInFlow = (el: HTMLElement, parent: HTMLElement) => {
 
   if (style.overflow && style.overflow !== 'visible') return;
   if (parentStyle.float !== 'none') return;
+  if (parent && parentStyle.display === 'grid') {
+    return;
+  }
   if (
     parent &&
     parentStyle.display === 'flex' &&


packages/utils/src/getDOMInfo.ts
```diff
@@ -55,6 +55,9 @@ export const styleInFlow = (el: HTMLElement, parent: HTMLElement) => {
 
   if (style.overflow && style.overflow !== 'visible') return;
   if (parentStyle.float !== 'none') return;
+  if (parent && parentStyle.display === 'grid') {
```
timc1, line 58: Trying to understand the mechanism under the hood for this logic. Why does `flex` specifically check for `flex-direction: column`? 🤔

prevwong: One of the oldest part of Craft 😄 So this part is related to the placement logic and it's mostly borrowed from GrapesJS. 

Let's say we have 2 elements - `A` for the element we're dragging and `B` for the element we're hovering over.

If element B is in "flow" - then we would only consider B's y-axis threshold relative to the current cursor position to determine if A should be dropped before or after B. This is because some elements are arranged vertically, and so naturally when you're gonna drop something at these vertically arranged elements, you would move the cursor from top-to-bottom rather than from left-to-right.

For example, if a child element's parent is of `flex-direction: column`, it would mean that it takes up the entire width of its parent, and should be arranged vertically. 
> Related to: https://github.com/artf/grapesjs//issues/1626

timc1: Ah, so it would be different in the case of `grid` – even if it's `grid-auto-flow: row` we can't assume elements are all arranged vertically?

prevwong: Yea, we would probably require some additional heuristics

prevwong: @timc1 Shall we merge this one in as a temporary solution for grid layouts? Then, we can revisit this issue with a more complete solution later?

timc1: @prevwong let's do that 👍

prevwong (approved):

timc1:
@prevwong mind cutting a `0.2.0-alpha.14`? 👋

prevwong:
@timc1 Done! 🥳

PR #178, opened 2021-02-05
feat: rework handlers

prevwong:
- Improves and simplifies EventHandlers internal API
- Removes hacky and unstable subscriber logic


packages/core/src/events/isEventBlockedByDescendant.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * Check if a specified event is blocked by a child
+ * that's a descendant of the specified element
+ */
+export const isEventBlockedByDescendant = (e, eventName, el) => {
```
matdru, line 5: Whats the example use case for this? 🤔

packages/utils/src/EventHandlers/DerivedEventHandlers.ts
```diff
@@ -0,0 +1,39 @@
+import { EventHandlers } from './EventHandlers';
+import { EventHandlerConnectors, EventHandlerUpdates } from './interfaces';
+
+export abstract class DerivedEventHandlers<
```
matdru, line 4: Note to self: what do derived handlers do ? 🤔

packages/utils/src/EventHandlers/EventHandlers.ts
```diff
@@ -0,0 +1,147 @@
...
+    const bindedListener = (e: CraftDOMEvent<HTMLElementEventMap[K]>) => {
+      // Store initial Craft event value
+      if (!e.craft) {
+        e.craft = {
+          blockedEvents: {},
+          stopPropagation: () => {},
+        };
+      }
+
+      if (!isEventBlockedByDescendant(e, eventName, el)) {
```
matdru, line 62: i wonder if this would be better named `wasHandledByDescendant` ? 🤔

matdru: My rationale is that it's easier to reason about these events as `handled` or `consumed` rather than `blocked` since blocked makes me feel like they were prevented from occurring somewhat 😅

matdru:
I think it looks good! I just wish there were few more comments around to explain purpose of certain things 😅 It might make it easier for new contributors to follow what's happening

packages/utils/src/EventHandlers/DerivedEventHandlers.ts
prevwong: It's a way to create EventHandlers that depend on another EventHandler. The derived EventHandlers would have be able to easily create connectors that composites of the parents connectors. Additionally, when the parent handlers are disable/enabled, so will the derived handlers

This is the case with the LayerEventHandlers which depends on the CoreEventHandlers.

packages/utils/src/EventHandlers/EventHandlers.ts
prevwong: > blocked makes me feel like they were prevented from occurring somewhat 

That's because they're being blocked. The descendant calls `.craft.stopPropagation` which blocks the event from further propagating to an ancestor.

matdru: Yeah i get it, but it's the propagation that is stopped, not the event itself, thats why i found the name confusing

matdru (approved):

packages/utils/src/EventHandlers/EventHandlers.ts
prevwong: I'm not sure if I follow. If the propagation is stopped, the parent event should by all means not fire - which is what we're trying to emulate here, inside the parent event handler.

Issue #183, opened 2021-02-11
Is there a way to detect when a child component is incoming into a parent container? 


raphkhan:
Hi, to improve the UX of the drag & drop, I want to change my container's color when I drag an incoming component into it.

Is there a way to detect when a child component is incoming into a parent container?

ankri:
Currently not, but there is my Draft PR #37 that aims to add that functionality. I just need to find some time to finish it.

hugominas:
@raphkhan I was wondering the same the other day, then I noticed that the DOM changed on dragOver. I think that for the time being you can change the styles using the CSS class "draggedOver"

vazitron:
Same issue here: https://github.com/prevwong/craft.js/issues/124

PR #184, opened 2021-02-11
Added events.draggedOver part 2

### LINKED ISSUE #28: Add draggedOver to state.events

ankri:
This PR adds `draggedOver` to the `events`. It's almost the same as #37 but targets the `next` branch and the new EventHandlers, which also gives us a cleaner API. (#37 will be obsolete).

Since the new EventHandlers introduced a `Set` as the store for the EventHandlers, we can now store not only the id of the node we are dragging over, but all of its ancestors, too.

Closes #28, #37, #183

---

The only complaint we have with the new API is the name `draggedOver`. @prevwong suggested to re-use the `hovered` event. If an element is being hovered, and the `dragged` event stores at least one nodeId, we are dragging over an element. I prefer a more explicit API over an implicit one. As a counter argument Prev suggested to make use of the `EventHelpers`. 

I, too am not a fan of the name `draggedOver` but lack the creativity to come up with a better name, but I prefer the DX of having an explicit event for dragging over elements. My argument is, that the developer could discover the API with just using code completion and without having to have a look at the documentation of `hovered` or looking at a guide / example.

Maybe @matdru,  @mresposito or @timc1 can chime in and discus the API. Not only these guys, but everyone is welcome to leave a comment. 👍

### DIFF
--- a/packages/core/src/editor/actions.ts
+++ b/packages/core/src/editor/actions.ts
@@ -278,6 +278,7 @@ const Methods = (
       this.setNodeEvent('selected', null);
       this.setNodeEvent('hovered', null);
       this.setNodeEvent('dragged', null);
+      this.setNodeEvent('draggedOver', null);
       this.setIndicator(null);
     },
 
--- a/packages/core/src/editor/store.tsx
+++ b/packages/core/src/editor/store.tsx
@@ -16,6 +16,7 @@ export const editorInitialState: EditorState = {
     dragged: new Set<NodeId>(),
     selected: new Set<NodeId>(),
     hovered: new Set<NodeId>(),
+    draggedOver: new Set<NodeId>(),
   },
   indicator: null,
   handlers: null,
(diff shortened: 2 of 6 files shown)


ankri:
Bad bot!

If I had known that kodiak would automerge I would have opened a draft PR. Sorry Prev.

ankri:
Here is the [example](https://deploy-preview-184--craftjs.netlify.app/examples/basic/)

matdru:
Huh, that's weird, kodiak should only attempt to merge once there is an approval for PR from at least one reviewer 🤔 That being said, i'm not a fan of automatic merges at all 😒

prevwong:
Yikes! The bot requires branch permissions to work correctly, but then we will lose the ability to auto release after merging. Will disable kodiak for now until we come up with a better CI solution 😄 

@ankri could you create a new PR with this branch again? Sorry for the troubles, the feature looks pretty good though!

fifi98:
Hey @prevwong, I noticed this commit has been reverted and the `draggedOver` event is not there anymore. What was the reason it got reverted?

PR #186, opened 2021-02-12
Added events.draggedOver part 2.5

### LINKED ISSUE #28: Add draggedOver to state.events

### LINKED ISSUE #183: Is there a way to detect when a child component is incoming into a parent container?

ankri:
> Kodak bot went crazy with #184 and automatically merged the PR. Hopefully this won't be auto merged. 

This PR adds `draggedOver` to the `events`. It's almost the same as #37 but targets the `next` branch and the new EventHandlers, which also gives us a cleaner API. (#37 will be obsolete).

Since the new EventHandlers introduced a `Set` as the store for the EventHandlers, we can now store not only the id of the node we are dragging over, but all of its ancestors, too.

Closes #28
Closes #37
Closes #183

---

The only complaint we have with the new API is the name `draggedOver`. @prevwong suggested to re-use the `hovered` event. If an element is being hovered, and the `dragged` event stores at least one nodeId, we are dragging over an element. I prefer a more explicit API over an implicit one. As a counter argument Prev suggested to make use of the `EventHelpers`. 

I, too am not a fan of the name `draggedOver` but lack the creativity to come up with a better name, but I prefer the DX of having an explicit event for dragging over elements. My argument is, that the developer could discover the API with just using code completion and without having to have a look at the documentation of `hovered` or looking at a guide / example.

Maybe @matdru,  @mresposito or @timc1 can chime in and discus the API. Not only these guys, but everyone is welcome to leave a comment. 👍


ankri:
The [example](https://deploy-preview-186--craftjs.netlify.app/examples/basic/)

packages/core/src/events/DefaultEventHandlers.ts
```diff
@@ -148,6 +148,9 @@ export class DefaultEventHandlers<O = {}> extends CoreEventHandlers<
               return;
             }
 
+            const ancestors = store.query.node(targetId).ancestors();
+            store.actions.setNodeEvent('draggedOver', [targetId, ...ancestors]);
+
```
prevwong, line 153: I'm wondering if we really need to include the ancestors here. Could we get away with just the `targetId` being the Node that's being dragged over? 

Then in the use case where we need to know if the current ancestor element contains the Node that's being dragged over, we could just do:

```tsx
// Inside a parent Node
const descendants = query.node(id).descendants();
// Not a nice api, but maybe a helper could be used:
const draggedOverDescendant = Array.from(state.events.draggedOver).find(id => descendants.includes(id))
```

ankri: I prefer having access to all the node ids that we drag over. But it's not a strong opinion, so I don't mind changing the API.

ankri:
@DavraYoung as promised. Thanks again for reminding me, that this still open

packages/core/src/events/DefaultEventHandlers.ts
```diff
@@ -150,6 +150,8 @@ export class DefaultEventHandlers<O = {}> extends CoreEventHandlers<
               return;
             }
 
+            store.actions.setNodeEvent('draggedOver', targetId);
+
```
prevwong, line 154: Could we possibly use `hovered` instead of having to add a `draggedOver` property to the state?

Then in the `getDraggedOverNodes` helper, we could do something like:
```
const hovered = state.events.hovered;
const dragged = state.events.dragged;

// If there's no nodes being dragged -
// then any hovered nodes are just being "regularly" hovered
if (dragged.size() === 0 || hovered.size() === 0) {
   return new Set();
}

// Might want to filter out the dragged node from the hovered set
return hovered;
```

ankri: I finally took up the PR again and I really like your suggestion. There are a few problems though :'(

Currently `hovered` only updates on `mouseover`. For this suggestion to work, I would need to also add `store.actions.setNodeEvent('hovered', targetId);` to `dragenter` since `mouseover` is not triggered when dragging.

One could consider this a breaking change because now `hover` would be true when dragging nodes and current users would need to update their `useNode`s to only show their hovered state when we're not dragging.

But the biggest problem would be that `state.events.dragged` does nodes from `connectors.create`. And sadly cannot add `actions.setEvent('dragged', tree.rootNodeId)` to the `dragstart` of `connectors.create` because the tree has not yet been added.

---

To recap: We could leverage `hovered` and `dragged` when we A) add a breaking change regarding hovered and B) solve the problem of finding out whether a node is being created with `connectors.create` right now.

ankri: Okay. I found a solution for the second problem. We can use the `indicator` instead of `dragged`

```
    getDraggedOverNodes(): Set<NodeId> {
      const hoveredNodes = state.events.hovered;

      if (hoveredNodes.size === 0 || state.indicator === null) {
        return new Set();
      } else {
        const hoveredId = Array.from(hoveredNodes)[0];

        return new Set([
          hoveredId,
          ...NodeHelpers(state, hoveredId).ancestors(),
        ]);
      }
    },
```

prevwong: Hm I see, let's maybe try it out and see how it works with the landing example (because there we make use of the `hovered` state to display the highlighted borders).

Otherwise, yeah maybe a new event `draggedOver` might be the way to go; still gotta warm up to that name though.

linxianxi:
Is there any progress？

ankri:
We haven't decided on a name, since both @prevwong  and I don't really like the name `draggedOver`

michaeled:
Hey @ankri, when might this be included in a release?

ankri:
Closing in favor of #374

ankri:
Closing due to no time available

PR #187, opened 2021-02-12
chore: replace staging basic example URL with relative URL

ankri:
Currently when netlify is hosting the PR preview the URL for the basic example will link to the production example. This should fix that by providing a relative link.

Would it be possible to configure Netlify in a way, that they not only post the URL of the landing example, but also the basic example in their comment?

### DIFF
--- a/config.js
+++ b/config.js
@@ -6,8 +6,8 @@ module.exports = {
       BASIC_EXAMPLE: 'https://craft.js.org/examples/basic/',
     },
     staging: {
-      LANDING: 'https://craftjs.netlify.com/',
-      DOCUMENTATION: 'https://craftjs.netlify.com/r/docs/overview/',
+      LANDING: '/',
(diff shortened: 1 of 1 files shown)


config.js
```diff
@@ -8,7 +8,7 @@ module.exports = {
     staging: {
       LANDING: 'https://craftjs.netlify.com/',
       DOCUMENTATION: 'https://craftjs.netlify.com/r/docs/overview/',
```
prevwong, line 10: Can we change these too?

prevwong (changes requested):
Thanks!

config.js
ankri: Done 👍

prevwong (approved):

prevwong:
Closing since this is not needed anymore since we refactored our docusaurus site structure

Issue #180, opened 2021-02-09
Ability to select a node immediately after connectors.create()


timc1:
Trying to figure out a way to select a node after dragging from the toolbox into the canvas. It doesn't look like there is a way to get the fresh `nodeId` when using ` connectors.create(ref, Component)`

I might be overlooking something simple, but lmk!

GuglaniNiharika:
@timc1 https://github.com/prevwong/craft.js/pull/22
In your component, you can call the selectNode function with nodeId. 
```
const { actions: {selectNode}, } = useEditor();
const {
    id,
    actions: { setProp },
  } = useNode();

  useEffect(() => {
    selectNode(id);
  }, []);
```
This will auto-select your node after dragging it from the toolbox into the canvas

timc1:
@GuglaniNiharika if we were to add that for every User component, that would effectively call `selectNode` on every User component's initial mount?

GuglaniNiharika:
@timc1 Yes, it will. So, if you drag and drop some component say `Heading` into canvas, it would make `selected` property of that particular node as true

timc1:
Yeah, makes sense – the part that is concerning is that every User component on the screen will call `selectNode` on the initial mount 🤷

GuglaniNiharika:
@timc1 You want to select the recent most dropped component right? So, if you have the above code in your user components, the initial mount will only happen for the component you are dropping and the id which you receive from useNode will be the id of the recent most instance of the user component. 
So, if your canvas has 2 instances of user component A and 1 instance of user component B, and now you try and drop user component A again into the canvas, the useNode will give you the id of this last instance. selectNode will be called only with this last instance's id.  And user component B won't be rerendered

addlistener:
```
      const {parentNodeId, idx} = props.getParentAndIdx(state, query);
      console.log({parentNodeId, idx});
      const nodeTree = query.parseReactElement(el as ReactElement).toNodeTree();
      // console.log('add to canvas');
      actions.addNodeTree(nodeTree, parentNodeId, idx);

      setTimeout(() => {
          const nodeId = query.node(parentNodeId).descendants()[idx]!;
          actions.selectNode(nodeId);
      }, 100);
```

Here's sth I came up with. But the problem is that `query.node(...)` does not get the latest tree structure. Still trying

abird:
[...]

PR #188, opened 2021-02-12
feat: Added a callback function to connectors.create to get access to the added node

ankri:
With this PR #180 should be possible without the `useEffect` workaround.

It is now possible to use the following code:

```jsx
<MaterialButton
  ref={(ref) => {
    connectors.create(
      ref,
      <Button text="Click me" size="small" />,
      // this will be called, when the node is dropped
      (nodeTree) => {
        // we can now select the node or do other stuff with it
        actions.selectNode(nodeTree.rootNodeId);
      }
    );
  }}
  variant="contained"
>
  Button
</MaterialButton>
```

---

To add this feature it was necessary to modify the `ConnectorRegistry`. I didn't want to change too much, that's why I added a `additionalArguments` attribute to the `ConnectorToRegistry` type and object. It would be possible to merge `opts` and `additionalArguments` and spread it instead of using [both](https://github.com/prevwong/craft.js/compare/next...ankri:connectors-create-callback?expand=1#diff-de010ce279171772b23f05f02efbdffcbe517cb55c59f41208355ccc8487ac1aR59-R60) .

I only added a callback function to `connectors.create` since I couldn't come up with a use case for the other connectors, but I think we could add such a callback to the others, too.

### DIFF
--- a/packages/core/src/events/CoreEventHandlers.ts
+++ b/packages/core/src/events/CoreEventHandlers.ts
@@ -3,6 +3,10 @@ import { DerivedEventHandlers, EventHandlers } from '@craftjs/utils';
 import { EditorStore } from '../editor/store';
 import { NodeId, NodeTree } from '../interfaces/nodes';
 
+export interface CreateHandlerOptions {
+  onCreate: (nodeTree: NodeTree) => void;
+}
+
 export class CoreEventHandlers<O = {}> extends EventHandlers<
   { store: EditorStore } & O
 > {
@@ -16,7 +20,7 @@ export class CoreEventHandlers<O = {}> extends EventHandlers<
       create: (
         el: HTMLElement,
         UserElement: React.ReactElement,
-        callbackFunction?: (nodeTree: NodeTree) => void
+        options?: Partial<CreateHandlerOptions>
       ) => {},
     };
   }
--- a/packages/core/src/events/CoreEventHandlers.ts
+++ b/packages/core/src/events/CoreEventHandlers.ts
@@ -1,7 +1,7 @@
 import { DerivedEventHandlers, EventHandlers } from '@craftjs/utils';
 
 import { EditorStore } from '../editor/store';
-import { NodeId } from '../interfaces/nodes';
+import { NodeId, NodeTree } from '../interfaces/nodes';
 
 export class CoreEventHandlers<O = {}> extends EventHandlers<
   { store: EditorStore } & O
@@ -13,7 +13,11 @@ export class CoreEventHandlers<O = {}> extends EventHandlers<
       hover: (el: HTMLElement, id: NodeId) => {},
       drag: (el: HTMLElement, id: NodeId) => {},
       drop: (el: HTMLElement, id: NodeId) => {},
-      create: (el: HTMLElement, UserElement: React.ReactElement) => {},
+      create: (
+        el: HTMLElement,
+        UserElement: React.ReactElement,
+        callbackFunction?: (nodeTree: NodeTree) => void
+      ) => {},
     };
   }
 }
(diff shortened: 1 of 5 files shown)


packages/utils/src/EventHandlers/EventHandlers.ts
```diff
@@ -77,10 +77,11 @@ export abstract class EventHandlers<O extends Record<string, any> = {}> {
     return Object.keys(connectors).reduce(
       (accum, connectorName) => ({
         ...accum,
-        [connectorName]: (el, opts) => {
+        [connectorName]: (el, opts, ...args) => {
           this.registry.register(el, {
             opts,
             name: connectorName,
+            additionalArguments: args,
             connector: connectors[connectorName],
```
prevwong, line 85: Maybe we could change this a little. The first argument remains as `el` the DOM element, the second element becomes the required value, and then a 3rd argument for "options". So:

```suggestion
        [connectorName]: (el, required, opts) => {
          this.registry.register(el, {
            name: connectorName,
            required,
            opts,
            connector: connectors[connectorName],
```

This is related to #152 where we discussed the ability to add additional values as the 3rd parameter.

packages/core/src/events/CoreEventHandlers.ts
```diff
@@ -13,7 +13,11 @@ export class CoreEventHandlers<O = {}> extends EventHandlers<
       hover: (el: HTMLElement, id: NodeId) => {},
       drag: (el: HTMLElement, id: NodeId) => {},
       drop: (el: HTMLElement, id: NodeId) => {},
-      create: (el: HTMLElement, UserElement: React.ReactElement) => {},
+      create: (
+        el: HTMLElement,
+        UserElement: React.ReactElement,
+        callbackFunction?: (nodeTree: NodeTree) => void
```
prevwong, line 19: In this case, the required value will be the `UserElement`; where as the 3rd parameter will be the `options` object. For example:

```suggestion
      create: (
        el: HTMLElement,
        UserElement: React.ReactElement,
        options?: { onCreate: (nodeTree: NodeTree) => void }
```

prevwong (changes requested):

packages/utils/src/EventHandlers/EventHandlers.ts
ankri: Totally forgot about #152. Makes sense to make `opts` and object.

packages/core/src/events/CoreEventHandlers.ts
ankri: So with #152 in mind, options should be typed:
```
interface OnCreateOptions {
  onCreate?: (nodeTree: NodeTree) => void;
  previewImage?: string;
}
```

prevwong: Can we do `onCreate` first? Then we can redo the `previewImage` feature in the next PR? Just so we can keep things atomic and easily reversible 😅

ankri: Yes! That was my plan, but poorly communicated on my part

ankri:
I added the suggested changes. And paved the way to solve #152 at a later time.

packages/utils/src/EventHandlers/ConnectorRegistry.ts
```diff
@@ -5,13 +5,13 @@ import { Connector } from './interfaces';
 
 type ConnectorToRegister = {
   name: string;
-  opts: Record<string, any>;
-  additionalArguments: any[];
+  required: Record<string, any>;
```
prevwong, line 8: I think this should just be `any` now

ankri: done

prevwong (approved):

nicosh:
is this function avaible in 0.1.0-beta.16 or only in 0.2.0-alpha.21?

Issue #42, opened 2020-02-28
How to export the page as html


DanyRupes:
I don't know how to export my page

prevwong:
Unfortunately, this is currently out of our scope. For now, you can only serialize the editor state into JSON, and pass the JSON to the editor to reload the state. If you still need to export to HTML, you'll likely need to fiddle around with the JSON output and create your own HTML code generator.

jasonadkison:
You could also build a separate version of your Editor and user components that render the editor state as "read only" and then select the `element.outerHTML` of the rendered DOM node you want.

Enva2712:
The [Quill editor](https://quilljs.com/) runs into similar problems when trying to render static markup. Luckily, React saves the day here with the [ReactDOMServer.renderToStaticMarkup()](https://reactjs.org/docs/react-dom-server.html#rendertostaticmarkup) method.

chungchi300:
@Enva2712 @jasonadkison Can you guys provide some sample code, please? It will be very helpful for this key common requirement.

Enva2712:
Sure, here's an example function that takes the editor's exported state and returns markup. It depends on the user components that the editor was given to produce the state.
```jsx
import ReactDOMServer from 'react-dom/server';
import { Editor, Frame } from '@craftjs/core';
import userComponents from './user-components';

function renderMarkup(JSONStateString) {
  return ReactDOMServer.renderToStaticMarkup(<Editor enabled={false} resolver={userComponents}>
    <Frame json={JSONStateString} />
  </Editor>);
}
```

chungchi300:
@Enva2712 I have tried this but not lucks using the official landing page example, it returns an empty string
```javascript
{
[...]

PR #189, opened 2021-02-13
fix: handlers enabled when options.enabled=false

### LINKED ISSUE #181: Nodes are still draggable when options.enabled is false
**Describe the bug**

We're able to drag around a node that is connected when passing `enabled={false}` to the Craft provider.

To bypass this, we can ensure to not attach the drag handler when not enabled, but this should work regardless `<div ref => enabled && connect(drag(ref))>...</div>`

prevwong:
Fix #181

### DIFF
--- a/packages/core/src/events/Events.tsx
+++ b/packages/core/src/events/Events.tsx
@@ -52,17 +52,17 @@ export const Events: React.FC = ({ children }) => {
 
   // Disable/Enable handlers when the enabled state is toggled
   useEffect(() => {
-    if (!handlersRef.current) {
+    if (!handlers) {
       return;
     }
 
     if (!enabled) {
-      handlersRef.current.disable();
+      handlers.disable();
       return;
     }
 
-    handlersRef.current.enable();
-  }, [enabled]);
+    handlers.enable();
+  }, [enabled, handlers]);
 
   return handlers ? (
     <EventHandlerContext.Provider value={handlers}>
--- a/packages/utils/src/EventHandlers/ConnectorRegistry.ts
+++ b/packages/utils/src/EventHandlers/ConnectorRegistry.ts
@@ -53,6 +53,10 @@ export class ConnectorRegistry {
     this.registry.set(this.getConnectorId(element, toRegister.name), {
       opts: toRegister.opts,
       enable: () => {
+        if (cleanup) {
+          cleanup();
+        }
+
         cleanup = toRegister.connector(element, toRegister.opts);
       },
       disable: () => {
--- a/packages/utils/src/EventHandlers/tests/EventHandlers.test.ts
+++ b/packages/utils/src/EventHandlers/tests/EventHandlers.test.ts
@@ -114,9 +114,9 @@ describe('EventHandlers', () => {
       instance.enable();
     });
 
-    it('should only re-attach init', () => {
+    it('should cleanup existing and re-attach init', () => {
       Object.keys(instance.connectors).forEach((key) => {
-        expect(handlers[key].cleanup).toHaveBeenCalledTimes(0);
+        expect(handlers[key].cleanup).toHaveBeenCalledTimes(1);
         expect(handlers[key].init).toHaveBeenCalledTimes(1);
       });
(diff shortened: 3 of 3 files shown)


packages/core/src/events/Events.tsx
```diff
@@ -52,17 +52,17 @@ export const Events: React.FC = ({ children }) => {
...
     if (!enabled) {
-      handlersRef.current.disable();
+      handlers.disable();
       return;
     }
 
-    handlersRef.current.enable();
-  }, [enabled]);
+    handlers.enable();
+  }, [enabled, handlers]);
```
ankri, line 65: I think it would be enough to add `handlersRef.current` to the dependency array:

```tsx
  // Disable/Enable handlers when the enabled state is toggled
  useEffect(() => {
    if (!handlersRef.current) {
      return;
    }

    if (!enabled) {
      handlersRef.current.disable();
      return;
    }

    handlersRef.current.enable();
  }, [enabled, handlersRef.current]);
```

ankri: My tests seem to confirm, that it is enough to add `handlersRef.current` to the dependency array

prevwong: Is it good practice though? I don't think we should depend on ref values as dependencies? https://github.com/facebook/react/issues/16121

ankri: Wouldn't that be the case for `useMemo` then?

```tsx
  const memoizedStore = React.useMemo(() => store, [store]);
  const memoizedHandlers = React.useMemo(() => handlers, [handlers]);

  // Generate new handlers whenever the editor deserializes a new state
  useEffect(() => {
    // TODO: Remove this when we refactor the EditorState
    actions.history
      .ignore()
      .setState((state) => (state.handlers = handlersFactory(memoizedStore)));

    return () => {
      if (!memoizedHandlers) {
        return;
      }

      memoizedHandlers.cleanup();
    };
  }, [actions, handlersFactory, hydrationTimestamp]);

  // Disable/Enable handlers when the enabled state is toggled
  useEffect(() => {
    if (!memoizedHandlers) {
      return;
    }

    if (!enabled) {
      memoizedHandlers.disable();
      return;
    }

    memoizedHandlers.enable();
  }, [enabled, memoizedHandlers]);
```

But... why do we need to use `ref` or `memo` anyways? 🤔 

Since we do
```tsx
  const storeRef = useRef(store);
  storeRef.current = store;
  const handlersRef = useRef(handlers);
  handlersRef.current = handlers;
```
and I cannot find a place where we would need to have access to "reffed" version of the store or handlers? Am I missing something? 🤔

prevwong: Yeah I need the ref values here: https://github.com/prevwong/craft.js/blob/e250f449841feb007c4e54448dbdc8b57d39bba0/packages/core/src/events/Events.tsx#L41-L49

If I use `memoizedStore` and `memoizedHandlers` like in your example - then I will need to add those as a dependency to that `useEffect` hook, but that will cause that hook to execute everytime `handlers` and `store` is updated.

ankri: Thanks. LGTM then 🙂 👍

timc1:
@prevwong can we get a release with this? 💯

prevwong:
@timc1 Done! Sorry about the delay; really gotta fix the CI for the next branch 😁

timc1:
@prevwong this seems to still be the case when we have nodes like so:

```
<Element is={SomeComponent} canvas>
  <ChildBlock />
  <ChildBlock />
  <ChildBlock />
</Element>
```

wherein the `ChildBlock`s show the handlers when the editor is disabled 😞

hugominas:
Same here, the latest working release seems to be 0.2.0-alpha.13 
I was thinking about creating a separate renderer like what was discussed in this issue https://github.com/prevwong/craft.js/issues/42

timc1:
Got around this by placing the drag listener on a small drag handle, which itself is not rendered when the editor is disabled

PR #191, opened 2021-02-13
Fix: Updating props from the outside

ankri:
My attempt at fixing the bug described in #190

You can test it in the [basic example](https://deploy-preview-191--craftjs.netlify.app/examples/basic/)

I found a solution but hit a dead end. With this fix the props of the components get updated. *BUT* the component now will be rendered twice. Once with the old props and immediately with the new props.

Maybe someone has an idea?

### DIFF
--- a/packages/core/src/nodes/Element.tsx
+++ b/packages/core/src/nodes/Element.tsx
@@ -1,5 +1,5 @@
 import { ERROR_TOP_LEVEL_ELEMENT_NO_ID, useEffectOnce } from '@craftjs/utils';
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import invariant from 'tiny-invariant';
 
 import { NodeElement } from './NodeElement';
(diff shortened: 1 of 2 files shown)


timc1:
Something like this would be wonderful – it’s currently a bit of a pain to try and update a component with a global settings provider; for instance, trying to set some global colors and themes, we’ve had to place the hooks inside the blocks themselves. Ex; accessing `useContext(PropertiesContext)` within a `Text` as passing props to `Text` will only seed the block on the initial mount.

matdru:
I'm not really sure this is the correct way going about this, this essentially results in any extra props getting automatically copy pasted into Craft state, which might result in a lot of accidental props ending up in the state, it also mixes craft state with react state and eliminates a single source of truth for whats being rendered. I think to update props like that they should be injected straight into the state, rather than passed down as a prop and picked up by effect.

timc1:
@matdru do you have a recommended solution for handling something like global colors? For instance, a number of User blocks that inherit from a certain value in a provider. Changes in the provider would effectively update the user blocks?

real-ashwin:
@ankri Is there any plans to merge this PR or fix this issue in a different way?

PR #202, opened 2021-02-25
Fix npm install failure on landing example

xQwexx:
Fix npm install failure on landing example
And update landing example dependencies
It solve the Console warning: #203


prevwong (changes requested):
Thank you for this PR @xQwexx! 

I believe this PR is trying to couple of unrelated things (ie: updating dependencies, UI changes and the LayerNode bug). 

However, I would prefer if we could keep the changes atomic so that it's easier to revert a PR if something does go wrong. So if we could limit this PR to just solve a single issue (ie: the LayerNode bug in #203 and remove the other changes) - that would be great!

xQwexx:
Thank you for the review @prevwong 

This PR contains the updating of the dependencies, because of the change in the updated dependencies syntaxes, there was a need some UI changes
This LayerNode bug also showing because of the updated dependency (it also a really small change) so that is the reason I added to this pr but i can move that one for another pr

prevwong:
@xQwexx If we just kept the changes made to `LayerNode`, does that fix the original issue? If so, then let's keep this PR to only contain changes made to `LayerNode`.

Otherwise, if we absolutely need to update the dependencies to resolve the original issue - which dependencies that needs to be updated to fix the issue? I doubt most of these dependencies updated in this PR plays a role (ie: `babel-plugin-inline-react-svg`, `next`, `next-seo` or `react-contenteditable`). If it's `react`/`react-dom` that needs to be updated, we need to know what changed in those packages that required us to update it, so that we have a clear understanding for doing so.

xQwexx:
@prevwong I updated the dependencies because there was a lot of deprecated dependency.
Also I think the examples dependencies should be updated for time to time, 
because I feel the people use it as starting point

prevwong:
Closing in favour of #246. 

Thank you for opening this PR nevertheless! 💯

PR #212, opened 2021-03-10
feat: additional indicator options

timc1:
This PR enables us to add additional indicator options, specifically controlling:

- the ability to override the default `0.2ms` transition
- the thickness of the rendered indicator

https://user-images.githubusercontent.com/12195101/110573151-c04ed980-8128-11eb-8d60-0c06eac6473a.mp4

### DIFF
--- a/cypress/integration/history.ts
+++ b/cypress/integration/history.ts
@@ -51,7 +51,7 @@ describe('Test History', () => {
       .should('have.length', 1);
 
     // TODO currently I do not know why but without the wait. test gets flaky
-    //      and fails randomly because cypress is not able to drag the card into the container 
+    //      and fails randomly because cypress is not able to drag the card into the container
     cy.wait(100);
     cy.getByTestId('frame-card')
       .as('card')
--- a/packages/core/src/events/Events.tsx
+++ b/packages/core/src/events/Events.tsx
@@ -73,12 +73,13 @@ export const Events: React.FC = ({ children }) => {
               indicator.placement,
               getDOMInfo(indicator.placement.parent.dom),
               indicator.placement.currentNode &&
-                getDOMInfo(indicator.placement.currentNode.dom)
+                getDOMInfo(indicator.placement.currentNode.dom),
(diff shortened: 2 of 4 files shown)


cypress/integration/history.ts
```diff
@@ -51,7 +51,7 @@ describe('Test History', () => {
       .should('have.length', 1);
 
     // TODO currently I do not know why but without the wait. test gets flaky
-    //      and fails randomly because cypress is not able to drag the card into the container 
+    //      and fails randomly because cypress is not able to drag the card into the container
```
timc1, line 54: 🤷

packages/utils/src/RenderIndicator.tsx
```diff
@@ -1,9 +1,14 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 
-export const RenderIndicator: React.FC<any> = ({ style, parentDom }) => {
+export const RenderIndicator: React.FC<any> = ({
+  style,
+  parentDom,
+  ...rest
```
timc1, line 7: Can be more specific than a spread, no major thoughts here.

prevwong: Do we really need `...rest` here? I don't think there's a use case for spreading the props yet.

cypress/integration/history.ts
prevwong: Yeah, Cypress tests can be a little flaky at the moment since there really isn't a proper/native way of handling events (especially dnd events) yet and the best we can do for now is to emulate Craft's dnd behaviour to perform tests.

cc: @ankri

prevwong: Oh you were fixing the linting issue, sorry I thought you were asking about the flaky test comment 😆

packages/utils/src/RenderIndicator.tsx
timc1: Removed, it was an artifact from a previous idea 👍

prevwong (approved):

PR #221, opened 2021-03-19
Fix drag events not moving indicator when staying in the same node

nicoladefranceschi:
This pull request fixes a problem when dragging nodes.
Drag events are not moving indicator when the mouse is staying in the same node.
This gifs explain it better:

If the mouse moves on different nodes, it works as expected:
craftjs_bug_1

If you move the mouse **staying** in the same node, it doesn't work:
craftjs_bug_2


nicoladefranceschi:
Sorry, I'm closing this PR, I opened another one

PR #222, opened 2021-03-19
Fix drag events not moving indicator when staying in the same node

nicoladefranceschi:
This pull request fixes a problem when dragging nodes.
Drag events are not moving indicator when the mouse is staying in the same node.
This gifs explain it better:

If the mouse moves on different nodes, it works as expected:
craftjs_bug_1

If you move the mouse **staying** in the same node, it doesn't work:
craftjs_bug_2


packages/core/src/events/DefaultEventHandlers.ts
```diff
@@ -15,6 +17,41 @@ export class DefaultEventHandlers extends CoreEventHandlers {
...
+      x,
+      y,
+    });
+
+    if (!indicator) {
+      return;
+    }
+    this.store.actions.setIndicator(indicator);
+    DefaultEventHandlers.indicator = indicator;
+  }
```
prevwong, line 52: This part seems to be the same as the `dragenter` handler used in the `drop` connector. 

Can we rename this function to something like `computeIndicator`, so we can use for both the `dragover` and `dragenter` handler?

DavraYoung: any update on this?

hugominas:
Great work @nicoladefranceschi  this is looks like its solving a core bug that has been nagging us for a while any chance of resolving this confilets before the next release?

nicoladefranceschi:
@hugominas sorry I'm quite busy right now, maybe someone else can take over and make the necessary changes so that this can be merged?

ankri:
I think I fixed the merge conflicts and added @prevwong suggestion to rename `onDragOver` to `computeDragIndicator`.

But... I cannot tell the difference between the before https://craft.js.org and the new behavior. Did I mess up while resolving the merge conflict?

hugominas:
Hi @ankri thank you! To be honest after testing this PR locally I did not find any noticeable difference. I was facing a drag and dropping issue caused by the Elements not having enough space between them and I only realised this after. Probably @nicoladefranceschi can help us.

leandroruel:
all those PRs are pratically abandoned

PR #234, opened 2021-04-23
fix(docs): corrects various typos in project documentation

plan-do-break-fix:
Scope of changes is restricted to docs only


prevwong (approved):
Thank you! 💯

Issue #228, opened 2021-04-01
Stagnant linked nodes remain in json after parent is deleted.


Spenc84:
**Describe the bug**
When deleting a node that has linked nodes assigned to it, the node itself is removed from the json but it's linked nodes are not. 

**To Reproduce**
Create a node that renders a linked node and then delete it and check the resulting json. The original node will be gone but the linked node will remain in the json even though it no longer has a parent to render it.

**Expected behavior**
When a node is deleted it's linked nodes should be removed along with it's child nodes.

nicosh:
Maybe this can help #179

```
    // remove unused nodes
    const getDescendants = (parsedData, nodeId) => {
        const nodes = {}
        function recursiveMap(nodeId) {
            const node = parsedData[nodeId]
            node.nodes.forEach((nodeId) => recursiveMap(nodeId))
            Object.keys(node.linkedNodes).map((key) => {
                const nodeId = node.linkedNodes[key]
                return recursiveMap(nodeId)
            })
            nodes[nodeId] = node
        }
        recursiveMap(nodeId)
        return nodes
    }
```

Spenc84:
Thanks. I guess I can periodically run that on the root node until the issue can get sorted out.

prevwong:
Related to @179

Issue #179, opened 2021-02-09
actions.serialize() should return only the nodes that are necessary


timc1:
**Describe the bug**

The `EditorState` grows as more blocks are added to the canvas; however, when a block is deleted the linkedNodes still exist in the tree

**To Reproduce**

1. Add a few blocks and serialize the state of the editor
2. Delete all blocks – then serialize again
3. Notice the `ROOT` contains no `linkedNodes` and `nodes`, but the references to the old nodes still exist

<img width="402" alt="Screen Shot 2021-02-09 at 1 24 39 PM" src="https://user-images.githubusercontent.com/12195101/107409441-2caecc80-6ada-11eb-8d53-7d060352b9a9.png">

nicosh:
I have the same issue, and this is really annoying when you save data to db since you use a lot of unnecessary space. Also react hydration data is bigger  that needed.

timc1:
@nicosh I was able to get a sanitized version of what `serialize` returned like so  👍

```js
import { SerializedNode } from "@craftjs/core";

function getDescendants(
  parsedData: Record<string, SerializedNode>,
  nodeId: string
) {
  const nodes = {};

  function recursiveMap(nodeId) {
    const node = parsedData[nodeId];

    // Add Child nodes
    node.nodes.forEach((nodeId) => recursiveMap(nodeId));

    // Add linked nodes
    Object.keys(node.linkedNodes).map((key) => {
      const nodeId = node.linkedNodes[key];
      return recursiveMap(nodeId);
    });

    nodes[nodeId] = node;
  }
  recursiveMap(nodeId);
  return nodes;
}

/**
 * `sanitizeSerializedData`
 * Starting from a `rootId`, return only the necessary 
 * nodes by traversing child and linked nodes.
 */
export default function sanitizeSerializedData(
  serializedData: string,
  rootId: string
): Record<string, SerializedNode> {
  try {
    const parsed = JSON.parse(serializedData);
    const nodes = getDescendants(parsed, rootId);
    return nodes;
  } catch (error) {
    return {};
  }
}

```

prevwong:
Closing as this has been fixed! 💯

PR #236, opened 2021-04-26
fix: orphaned linked nodes

### LINKED ISSUE #176: Impossible to delete a node together with its linked nodes
It's impossible to delete a node together with its linked nodes.
If I have a node with id "A" which has a linked node with id "B", if I try to call actions.delete("A"), the linked node "B" still remains and I can't delete it after because its parent doesn't exist anymore.

prevwong:
Removes linked nodes when parent node is deleted

Fixes #176 #179 #228

### DIFF
--- a/packages/core/src/editor/actions.ts
+++ b/packages/core/src/editor/actions.ts
@@ -99,7 +99,7 @@ const Methods = (
     return parent;
   };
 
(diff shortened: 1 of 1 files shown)


prevwong:
Released in 0.1.0-beta.17 and 0.2.0-alpha.22! 🚀

PR #245, opened 2021-05-21
fix: linked node props

prevwong:
This PR removes the ability of synching `<Element />` props with its corresponding Linked Node, which was originally introduced in #48

## Rationale
- The existing behaviour introduces a bug whereby the props set in `<Element />` will take precendence over the values set in the corresponding Linked Node, which makes it impossible to persist changes made to the Linked Node.
```jsx
const Button = () => {
   return (
     <div>
       <Element id="header" is={Container} bg="#000" />
       {
         /** 
           // If the following Node state gets deserialised into the Editor at a later time,
           // The props set in the Node will be overridden by the props set in the above <Element />
          {
            type: Container,
            props: {
              bg: "#fff"
            }
          }
         **/
       }
    </div>                                                                                                                                      
}
```
-  The changes in this PR makes the behaviour of the `<Element />` component more consistent with the rest of Craft
   - Whereby once a Node is created for a given Element; the corresponding Node should be the single source of truth. In other words `<Element />` only exists as a means of configuring the initial state for a Node.


ankri (approved):
Great idea!

PR #251, opened 2021-06-01
Fixed the bug if `props` is undefined in recursion

slobber:
`props` maybe has a `type` property, but it isn't a craft `UserComponent`, so I recommend a double check of `prop.props`.

### DIFF
--- a/packages/core/src/utils/serializeNode.tsx
+++ b/packages/core/src/utils/serializeNode.tsx
@@ -31,7 +31,7 @@ export const serializeComp = (
         }
         return serializeComp(child, resolver);
       });
-    } else if (prop.type) {
+    } else if (prop.type && typeof prop.props === 'object') {
       result[key] = serializeComp(prop, resolver);
     } else {
       result[key] = prop;


packages/core/src/utils/serializeNode.tsx
```diff
@@ -31,7 +31,7 @@ export const serializeComp = (
         }
         return serializeComp(child, resolver);
       });
-    } else if (prop.type) {
+    } else if (prop.type && typeof prop.props === 'object') {
```
prevwong, line 34: The original intention here was to check if the current `prop` is a function and if so we will assume it's a User Component and attempt to serialise it as such. Perhaps, we could just do the following:

```suggestion
    } else if (typeof prop === 'function') {
```

tim-scanlin-ck: I have been running into this too, thanks for working on a fix! Wouldn't it be `typeof prop.type === 'function'` since the .type property has the component/function?

slobber:
it has been fixed

PR #253, opened 2021-06-06
chore(layers): mark styled-components as external dependencies

akingbadeayobami:
Basically I get this warning because of the fact there are two conflicting styled-components in the application

```
It looks like there are several instances of 'styled-components' initialized in this application. This may cause dynamic styles not rendering properly, errors happening during rehydration process and makes your application bigger without a good reason.
```

The changes made are as suggested on the styled-components docs
https://styled-components.com/docs/faqs#i-am-a-library-author-should-i-bundle-styledcomponents-with-my-library

### DIFF
--- a/packages/layers/README.md
+++ b/packages/layers/README.md
@@ -27,7 +27,7 @@
 ## Usage
 
 ```bash
-yarn add @craftjs/layers
+yarn add @craftjs/layers styled-components
 ```
 
 ```jsx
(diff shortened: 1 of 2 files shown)


prevwong:
Hi, thanks for opening this PR! Could you also update the README in the layers package so users would know to install the `styled-components` package as well:

```
yarn add @craftjs/layers styled-components
```

prevwong (approved):
Thank you! 💯

PR #256, opened 2021-06-11
Replaced lodash with lodash-es in web bundles

amkoehler:
While doing bundle analysis for our project I noticed that the use of lodash in `@craftjs/core` results in the entire `lodash` bundle included in the web bundle (results below). This likely became part of the bundle with the export of `testHelpers` in [this commit](https://github.com/prevwong/craft.js/commit/c47cacd9b40ebecd88133ff0f72407be0104a353). 

The `lodash` bundle is not tree shakeable, and in my experience even using a single lodash function results in the entire bundle getting included. The [recommended solution](https://developers.google.com/web/fundamentals/performance/optimizing-javascript/tree-shaking/#when_things_arent_so_straightforward) is to use `lodash-es` instead.

**Before** - using `lodash` imports
image

**After** - using `lodash-es` imports
image

- Updated all imports to use lodash-es instead of lodash
- Use jest file replacement for tests
- Added lodash dev dependency to support tests

### DIFF
--- a/packages/core/src/events/DefaultEventHandlers.ts
+++ b/packages/core/src/events/DefaultEventHandlers.ts
@@ -1,4 +1,4 @@
-import { isFunction } from 'lodash';
+import { isFunction } from 'lodash-es';
 
 import { CoreEventHandlers, CreateHandlerOptions } from './CoreEventHandlers';
 import { createShadow } from './createShadow';
--- a/package.json
+++ b/package.json
@@ -90,6 +90,9 @@
       "tsx",
       "js"
     ],
+    "moduleNameMapper": {
+      "^lodash-es$": "lodash"
+    },
     "modulePaths": [
       "<rootDir>/packages/core/"
     ],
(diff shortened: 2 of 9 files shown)


packages/core/src/events/DefaultEventHandlers.ts
```diff
@@ -1,4 +1,4 @@
-import { isFunction } from 'lodash';
+import { isFunction } from 'lodash-es';
```
prevwong, line 1: Hmm, I believe using the ESM import here would cause issues with our CJS bundle. Maybe we could use https://www.npmjs.com/package/babel-plugin-transform-rename-import to transform `lodash` imports to `lodash-es` for our ESM bundle?

amkoehler: Good catch. It looks like react-color did something similar: https://github.com/casesandberg/react-color/pull/749/files. I will look into it.

PR #271, opened 2021-07-07
fix: sync <Editor /> props with state options

### LINKED ISSUE #149: Cannot update editor options
This assignment to the function argument does not make sense, as it will do nothing.
https://github.com/prevwong/craft.js/blob/75285b20fe7fc37390c5b3bfd7b43719b0d022f8/packages/core/src/editor/Editor.tsx#L31

I noticed this, because I tried to toggle enabled on the editor, however with no effect.

aroramayur:
Closes #149

 - Updates Editor's options as a side-effect.

### DIFF
--- a/packages/core/src/editor/Editor.tsx
+++ b/packages/core/src/editor/Editor.tsx
@@ -1,5 +1,6 @@
 import { ERROR_RESOLVER_NOT_AN_OBJECT } from '@craftjs/utils';
-import React, { useEffect } from 'react';
+import { pickBy } from 'lodash';
+import React, { useEffect, useMemo } from 'react';
 import invariant from 'tiny-invariant';
 
 import { EditorContext } from './EditorContext';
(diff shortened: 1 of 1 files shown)


prevwong (approved):
Thank you! 💯

packages/core/src/editor/Editor.tsx
```diff
@@ -28,7 +28,7 @@ export const Editor: React.FC<Partial<Options>> = ({
   useEffect(() => {
     if (context && options)
       context.actions.setOptions((editorOptions) => {
-        editorOptions = options;
+        Object.assign(editorOptions, options);
       });
   }, [context, options]);
```
prevwong, line 33: Actually, on a second look, there's more to this - the `options` object changes way too frequently, thus the `useEffect` hook will get triggered too frequently as well.

On top of my head, I can think of 2 options:
1. Use `lodash/isEqualWith` to compare the `options` prop with the previous `options`. But I'm not a fan of using deepEquals, so it would be great if we can avoid this.

2. Memoize the `options` object:
```tsx
export const Editor: React.FC<Partial<Options>> = ({
  children,
  onNodesChange,
  onRender,
  enabled,
  resolver,
  indicator,
  handlers,
}) => {
  const options = useMemo(
    () => ({
      ...
    }),
    [onNodesChange, onRender, enabled, resolver, indicator, handlers]
  );

  ...
});
```

prevwong (changes requested):

prevwong (approved):
LGTM 💯

akucinskii:
Hey, was this pr released?

PR #272, opened 2021-07-07
feat: simplify EditorState

prevwong:
# Problem

The `EditorState` is largely denormalised and contains a bunch of redundant values. Initially, this was intended since the `EditorState` was directly exposed to all client APIs (ie: `useEditor` and `useNode`), hence having the EditorState to be denormalised allowed us to optimise for a better developer experience:

```tsx
useEditor(state => ({
   selectedNodeIds: state.events.selected,
   isRootNodeSelected: state.nodes[ROOT_NODE].events.selected,
   rootNodeDOM: state.nodes[ROOT_NODE].dom,
});
```

## Redundancy 
However, this redundancy meant that performing state mutations becomes redundant as well. For example, when we do `actions.setNodeEvent('selected', ROOT_NODE)`, we will need to update : 
- `state.events.selected.add(ROOT_NODE)` and 
- `state.nodes.ROOT_NODE.events.selected=true` 

## Node Complexity

Related to the redundancy point above, our current `Node` data structure is quite complex:
```tsx
type Node = {
  id: NodeId;
  data: {
      type: React.ElementType | string;
      name: string;
      displayName: string;
      props: Record<string, any>;
      custom: Record<string, any>;
      nodes: NodeId[];
      linkedNodes: Record<string, NodeId>;
      hidden: boolean;
      isCanvas: boolean;
  },
  dom: HTMLElement;
  related: Record<string, React.ElementType>;
  rules: NodeRules;
  events: Record<NodeEventTypes, boolean>;
}
```
However, most of these properties can be inferred from other parts of the editor and don't have to be apart of the data structure. For example, `node.events` can be inferred from `state.events`. Similarly, `data.type`, `related`, `rules` are all values that can be inferred from the `Resolver`.

The complexity of the current `Node` type results in difficulty in interacting with Nodes. For example, in order to create a new Node, you'll need to do so with `query.parseFreshNode` which returns the correct `rules`, `related` properties etc. 

## Fragility 
Another issue is that since the EditorState is directly tied to the client APIs, any change to the EditorState data structure would potentially result in a breaking change. For example, when we introduced multi-select where `state.events` was changed from `Record<NodeEventTypes, NodeId>` into `Record<NodeEventTypes, Set<NodeId>>`. This meant any codebase that was accessing `state.events` would essentially stop working correctly or y'know – crash 🔥 . 

# Solution

Simplifying `EditorState` would make Craft.js a lot more maintainable.

At the same time, we should strive to maintain the same level of (or better) developer experience. As you know - we have this concept of `query` methods that allowed developers to get useful information from the `EditorState`. In essence, the query methods was basically an abstraction layer on top of Craft's EditorState. Up until now, these methods were more like an optional or alternative way to interact with the state:

```tsx
useEditor((state, query) => ({
    isCanvas: state.nodes[ROOT_NODE].data.isCanvas,
    isCanvas: query.node(ROOT_NODE).isCanvas()
});
```

But if we made these query methods a first-class citizen in the client APIs, we could essentially achieve a comparable developer experience and we don't have as much to worry about when introducing breaking state changes.

```tsx
useEditor((query) => ({
    // even if the Node's isCanvas property is moved around, it wouldn't matter here
    isCanvas: query.node(ROOT_NODE).isCanvas()
});
```

This PR aims to simplify the EditorState while making query methods a first-class citizen in the client APIs.

### EditorState
```tsx
type EditorState = {
   nodes: Record<NodeId, Node>;
   events: Record<NodeEventTypes, Set<NodeId>>;
   enabled: boolean;
   timestamp: number;
   indicator: Indicator;
}
```

The main changes to `EditorState` is the removal of `options` which essentially contains configuration values. Most of these values never change and thus don't really need to be in the state. With the exception of `options.enabled` which has been moved up; all other option values has been moved out of the state and into the `EditorStore`'s config:

```tsx
class EditorStore {
   state: EditorState;
   config: {
       indicator: IndicatorOptions;
       handlers: CoreEventHandlers;
       resolver: Resolver;
   }
}
```

### Indicator
```tsx
type Indicator = {
    placement: {
         parentNodeId: NodeId;  // previously this was Node
         currentNodeId: NodeId | null; // previously this was Node | null
         index: number;
         where: 'before' | 'after';
    }
   error?: string;
}
```

### Node
```tsx
type Node = {
   id: NodeId;
   type: string;
   displayName: string;
   props: Record<string, any>;
   custom: Record<string, any>;
   isCanvas: boolean;
   hidden: boolean;
   nodes: NodeId[];
   linkedNodes: Record<string, NodeId>;
}
```

The biggest change in the `Node` structure is that `node.data` properties have been moved up in the data structure. The `type` property in Node is now always a string that references the component set in the `Resolver`; whereas previously it was `React.ElementType`:
```tsx
const resolver = {
   Container: MyContainerReactComponent
}

const nodeA = {
    type: 'Container',
    ...
}
```

The next biggest change with `Node` is the removal of transient or runtime values (ie: `events`, `related`, `rules` and `dom`). This is due to the fact that these values don't "define" what a Node actually is and can be stored and inferred from other parts of the editor:
- `events` can be inferred from `state.events` for the given `node.id`
- `rules` and `related` can be inferred from the React Component from the `Resolver` for the given `node.type`
- `dom` can be inferred from the `CoreEventHandlers` class; all DOM elements has been moved out the EditorState and is now stored in the EventHandlers:

Ultimately, the Node data structure will be a lot simpler:
```tsx
const node = { 
  id: 'node-a';
  type: 'Container',
  isCanvas: true,
  props: {...},
  custom: {...},
  hidden: false,
  nodes: [],
  linkedNodes: {},
}
```

Hence, we will no longer need additional data structures such as `FreshNode`, `SerializedNode`. Similarly, we will no longer need methods such as `parseFreshNode` or `parseSerializedNode`:
```tsx
// When adding Nodes, you can do so without needing to first perform parsing:
actions.add(node);

// previously:
// actions.add(query.parseFreshNode(legacyNode))
```

## Query
As mentioned earlier, these Query interfaces will be the primary way for interacting with the Editor from the client APIs. 

- `QueryMethods` has been refactored into `EditorQuery` which contains methods/accessors relating to the entire Editor
- `NodeHelpers` has been refactored into `NodeQuery` which contains methods/accessors relating to a `Node` in the state

```tsx
class EditorQuery {
   private readonly store: EditorStore;
 
   node(id: NodeId) {
      return new NodeHelpers(this.store, id);
   }

    isEnabled() {
       return this.store.getState().enabled;
    }

     // Methods from QueryMethod
    parseReactElement: (element: React.ReactElement) => { ... }
    parseFreshNode: () => {...}

    // Legacy state accessors:
    //  These are here to support existing APIs:: useEditor(state => state.nodes)
    get nodes() {
       return Object.keys(this.store.getState().nodes).map(nodeId => this.node(nodeId));
    }

    get events() {
        return this.store.getState().events;
    }

   get options() { ... }

    ... 
}

class NodeQuery {
    private readonly store: EditorStore;
    private readonly node: Node;
   
    get type() {
        return this.node.type;
    }

    getComponent() {
        return this.store.resolver[this.type];
    }

    getDOM() {
        return this.store.handlers.dom.get(this.node.id);
    }

    isSelected() {
        return this.store.query.event('selected').contains(this.node.id);
   }

   // ...all previous NodeHelpers methods
   isCanvas() {
        return this.node.isCanvas;
   }

   // Support legacy Node structure for existing APIs:: useEditor(state => state.nodes[ROOT_NODE].data)
   get data() {
       return {
           type: this.getComponent(), 
           ...
       }
   }

   ...
}
```

# Compatibility 

While this is a major change in the framework, most of these changes are backwards compatible:

```tsx
const { query } = useEditor((query: EditorQuery) => ({
   isRootSelected: query.root.isSelected(),
   someChildNodes: query.root.getChildNodes().map(nodeQuery => nodeQuery.id),
   rootDOM: query.root.getDOM(),
   someOtherNodeSelected: query.node('some-other-node').isSelected(),
   isEditorEnabled: query.isEnabled(),
   
   // all legacy stuff still works:
   isRootSelected: query.nodes[ROOT_NODE].events.selected,
   someChildNodes: query.nodes[ROOT_NODE].nodes,
   rootDOM: query.nodes[ROOT_NODE].dom,
   someOtherNodeSelected: query.nodes['some-other-node'].events.selected,
   isEditorEnabled: query.options.enabled,
});

// If you previously used the second `query` parameter, they will continue to work as well:
useEditor((state: EditorQuery, query: EditorQuery)) => {...}

// the following deprecated methods will continue to work as well:
const { actions , query } = useEditor();
query.serialize(); 
query.parseFreshNode();
query.parseSerializedNode();
actions.deserialize(...);
actions.setOptions(...);
```

```tsx
useNode((node: NodeQuery) => ({
    id: node.id,
    props: node.props,
    component: node.getComponent(),
    isSelected: node.isSelected(),
    dom: node.getDOM(),
    childNodesDOM: node.getNodes().map(nodeQuery => nodeQuery.getDOM()),
    
    // all legacy stuff still works:
    props: node.data.props,
    component: node.data.type,
    isSelected: node.events.selected,
    dom: node.dom,
    childNodesDOM: // ah ha! you can't do that before
});
```


hugominas:
This change is a move in the right direction. The state is bloated at the moment and this will surely help performance and clarity. Thank you for the time you put in this change @prevwong

jbbelcher53:
These changes seem very well thought out and important for the future maintainability of the library and the complexity of the biggest parts of it. Looking forward to implementing these changes and working with craft more. Thank you for the hard work!

packages/core/src/interfaces/store.ts
```diff
@@ -0,0 +1,219 @@
...
+} from './nodes';
+
+import { CoreEventHandlers } from '../events';
+import { EditorState, Resolver } from '../interfaces';
+import { RelatedComponents } from '../nodes/RelatedComponents';
+import { ActionMethods } from '../store/actions';
+
+export interface LegacyNodeQuery extends LegacyNode {
+  /**
+   * @deprecated
```
ankri, line 29: *Just using one of those `@deprecated` as an example`

Should the deprecated message contain a link to this PR / a page in the docs / a GitHub issue on how to migrate to the new syntax?

Or maybe print a [`deprecationWarning`](https://github.com/prevwong/craft.js/pull/272/files#diff-aaa1ec60726736a066bf50e9f34d4dc116028a30b488cd347f1a60922c8c953aR229) ?

packages/core/src/store/query/EventQueryImpl.ts
```diff
@@ -0,0 +1,71 @@
...
+  getState() {
+    return this.event;
+  }
+
+  getNodes() {
+    const nodes = Array.from(this.event);
+    return nodes.map((id) => this.store.query.node(id));
+  }
+
+  getNodeAtIndex(index: number) {
```
ankri, line 28: maybe check, that index is in range?

packages/core/src/store/query/EventQueryImpl.ts
```diff
@@ -0,0 +1,71 @@
...
+    const nodes = this.getNodes();
+    return nodes[index];
+  }
+
+  getFirst() {
+    return this.getNodeAtIndex(0);
+  }
+
+  getLast() {
+    return this.getNodeAtIndex(this.event.size - 1);
```
ankri, line 38: Would that not result in `-1` for a list of size 0? Maybe add a check as safe guard?

ankri (approved):
great PR! Thanks for your the detailed explanation and the hard work. I just have a suggestion regarding the deprecation and 2 minor nits.

packages/core/src/store/query/EventQueryImpl.ts
prevwong: Good catch

packages/core/src/interfaces/store.ts
prevwong: Will create a migration guide once this is a bit more stable

PR #281, opened 2021-07-26
feat: clear history

### LINKED ISSUE #242: Clear Builder History Programmatically
**Is your feature request related to a problem? Please describe.**
No it is not

**Describe the solution you'd like**
I looked at `actions.clearEvents()` but it wasn't what i thought it was, i think it would be really nice to be able to clear the builder history via a handy method, we do have `undo` and `redo` e.t.c

**Additional context**
None

prevwong:
Closes #242

### DIFF
--- a/packages/utils/src/useMethods.ts
+++ b/packages/utils/src/useMethods.ts
@@ -43,6 +43,7 @@ export type CallbacksFor<
       history: {
         undo: () => void;
         redo: () => void;
+        clear: () => void;
         throttle: (
           rate?: number
         ) => Delete<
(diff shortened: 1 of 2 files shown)


packages/utils/src/useMethods.ts
```diff
@@ -229,6 +230,12 @@ export function useMethods<
               case HISTORY_ACTIONS.REDO: {
                 return history.redo(draft);
               }
+              case HISTORY_ACTIONS.CLEAR: {
+                history.clear();
+                return {
+                  ...draft,
+                };
```
prevwong, line 237: Need to return a new copy of the state, otherwise collector methods will not be notified.

ankri (approved):

Issue #284, opened 2021-07-28
TypeError: r is not a function


throne1986:
Hi everyone,

I want to run the landing example using pure react js without typescript, so I have refactored the given example in this repo,

Here is my new repo: https://github.com/throne1986/landing-app

But, unfortunately, I keep getting the error: `TypeError: r is not a function

![127243905-c11b3d98-1360-4273-890e-ca68752ea058 1](https://user-images.githubusercontent.com/27904724/127295845-94d4a719-59c3-4a69-8960-a11f48f7ed45.png)

I am out of idea what I am doing wrong, can someone please tell me what is wrong here?

Thanks

ankri:
When I follow your [index.js](https://github.com/throne1986/landing-app/blob/master/src/index.js#L12) and your [App.js](https://github.com/throne1986/landing-app/blob/master/src/App.js#L9) then the [Viewport.js](https://github.com/throne1986/landing-app/blob/master/src/features/editor/Viewport/index.js)

You need to wrap your `<Viewport />` with a `<Editor />` component from `@craft.js/core` or remove the use of [`useEditor`](https://github.com/throne1986/landing-app/blob/master/src/features/editor/Viewport/index.js#L11-L13) since you are first using the `<Editor />` in your [Page.js](https://github.com/throne1986/landing-app/blob/master/src/features/pages/index.js#L32)

ankri:
In other words, you cannot use `useEditor` outside the context of the `<Editor />` component from the core package. 

I will try to open a PR that will produce a clearer error message for those cases.

throne1986:
> When I follow your [index.js](https://github.com/throne1986/landing-app/blob/master/src/index.js#L12) and your [App.js](https://github.com/throne1986/landing-app/blob/master/src/App.js#L9) then the [Viewport.js](https://github.com/throne1986/landing-app/blob/master/src/features/editor/Viewport/index.js)
> 
> You need to wrap your `<Viewport />` with a `<Editor />` component from `@craft.js/core` or remove the use of [`useEditor`](https://github.com/throne1986/landing-app/blob/master/src/features/editor/Viewport/index.js#L11-L13) since you are first using the `<Editor />` in your [Page.js](https://github.com/throne1986/landing-app/blob/master/src/features/pages/index.js#L32)

Thanks, bro that is true solved the error, but now I have another issue when I am rendering the Pages.js component it keeps rendering infinite 
[...]

PR #286, opened 2021-07-28
chore: add error when hooks is used outside of EditorContext

ankri:
Lately there were some issues resulting from users using `useEditor` and/or `useNode` outside of `<Editor />`.

This PR will throw an Error to warn the user.

Related:
#284

### DIFF
--- a/packages/core/src/editor/useInternalEditor.ts
+++ b/packages/core/src/editor/useInternalEditor.ts
@@ -38,10 +38,7 @@ export function useInternalEditor<C>(
   collector?: EditorCollector<C>
 ): useInternalEditorReturnType<C> {
   const store = useContext(EditorContext);
-  invariant(
-    Object.keys(store).length > 0,
-    ERROR_USE_EDITOR_OUTSIDE_OF_EDITOR_CONTEXT
-  );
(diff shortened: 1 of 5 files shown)


packages/core/src/editor/useInternalEditor.ts
```diff
@@ -35,8 +37,13 @@ export type useInternalEditorReturnType<C = null> = useCollectorReturnType<
 export function useInternalEditor<C>(
   collector?: EditorCollector<C>
 ): useInternalEditorReturnType<C> {
-  const handlers = useEventHandler();
   const store = useContext(EditorContext);
+  invariant(
+    Object.keys(store).length > 0,
```
prevwong, line 42: Why not `invariant(store, ERROR....)`?

ankri: The [default](https://github.com/prevwong/craft.js/blob/develop/packages/core/src/editor/EditorContext.tsx#L6) is an empty object. I didn't want to change that. That's why take the number of keys into account.

prevwong: Can we make it the default into `undefined` or `null`? Don't think that would change things too much.

ankri: You are right it did not change anything 👍

prevwong (approved):

Issue #268, opened 2021-07-05
Element‘s ID


963542298:
I want to know the rules for generating IDs，Thanks!

ankri:
Craft [uses](https://github.com/prevwong/craft.js/blob/develop/packages/core/package.json#L42) the package [shortid](https://www.npmjs.com/package/shortid) to generate the ids.

ankri:
By the way thanks for making me look into it, otherwise I wouldn't have seen the warning on the `shortid` page.

963542298:
Use Nano ID？

PR #287, opened 2021-07-28
chore: replace package shortid with nanoid

ankri:
Issue #268 made me look into [shortid](https://www.npmjs.com/package/shortid) and I stumbled upon the deprecation message and the safety warning.

This PR replaces `shortid` with the suggested [`nanoid`](https://www.npmjs.com/package/nanoid). For existing projects old NodeIDs should still work but new ones would be generated with `nanoid`.

Only "disadvantage" if you can call it that, is that ids generated by `nanoid` are longer than `shortid`'s ids.

### DIFF
--- a/packages/core/src/utils/getRandomNodeId.ts
+++ b/packages/core/src/utils/getRandomNodeId.ts
@@ -1,3 +1,0 @@
-import { nanoid } from 'nanoid';
-
-export const getRandomNodeId = nanoid;
--- a/packages/core/src/utils/getRandomNodeId.ts
+++ b/packages/core/src/utils/getRandomNodeId.ts
@@ -1,3 +1,3 @@
-import shortid from 'shortid';
+import { nanoid } from 'nanoid';
 
-export const getRandomNodeId = shortid;
+export const getRandomNodeId = nanoid;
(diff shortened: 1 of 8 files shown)


packages/core/src/utils/getRandomNodeId.ts
```diff
@@ -1,3 +1,3 @@
-import shortid from 'shortid';
+import { nanoid } from 'nanoid';
 
-export const getRandomNodeId = shortid;
+export const getRandomNodeId = nanoid;
```
prevwong, line 3: Would it be possible to get a shorter id? 

https://github.com/ai/nanoid/#custom-alphabet-or-size

packages/utils/src/EventHandlers/ConnectorRegistry.ts
```diff
@@ -30,7 +30,7 @@ export class ConnectorRegistry {
       return existingId;
     }
 
-    const newId = shortid();
+    const newId = nanoid();
```
prevwong, line 33: Let's rename `getRandomNodeId` to `getRandomId`, then we can import that utility method here too

ankri: To be able to do that, I moved nanoid to the `@craftjs/utils` package.

packages/core/src/utils/getRandomNodeId.ts
ankri: I shortened the id from 21 to 10. 7 (like shortid used - felt waaaay to short).

prevwong (approved):

PR #293, opened 2021-08-12
feat: throw error if component is missing from resolver

prevwong:
Prevents a common production error when a User Component is missing from resolver.

### DIFF
--- a/packages/core/src/editor/actions.ts
+++ b/packages/core/src/editor/actions.ts
@@ -55,7 +55,10 @@ const Methods = (
       if (typeof node.data.type !== 'string') {
         invariant(
           state.options.resolver[node.data.name],
-          ERROR_NOT_IN_RESOLVER
+          ERROR_NOT_IN_RESOLVER.replace(
+            '%node_type%',
+            `${(node.data.type as any).name}`
+          )
         );
       }
 
--- a/packages/core/src/editor/actions.ts
+++ b/packages/core/src/editor/actions.ts
@@ -8,6 +8,7 @@ import {
   ERROR_DELETE_TOP_LEVEL_NODE,
   CallbacksFor,
   Delete,
+  ERROR_NOT_IN_RESOLVER,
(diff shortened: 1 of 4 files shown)


ankri (changes requested):
I really love the idea of this PR, but I don't really like the error message. Because you cannot see which component is missing from the resolver.

My suggestion would be to add this code to `query.tsx`

```tsx
          invariant(
            name,
            ERROR_NOT_IN_RESOLVER.replace(
              '%node_type%',
              `${
                (node.data.type as any).name ||
                (node.data.type as any).displayName
              }`
            )
          );
```

and modify the `constants.ts`:
```tsx
export const ERROR_NOT_IN_RESOLVER =
  'The component type specified for this node (%node_type%) does not exist in the resolver';
```

---

But. Wouldn't it be better to throw an error in `resolveComponent.ts` instead?

```tsx
import { ERROR_NOT_IN_RESOLVER } from '@craftjs/utils';
import { Resolver } from '../interfaces';
import { Canvas } from '../nodes/Canvas';

export const resolveComponent = (
  resolver: Resolver,
  comp: React.ElementType | string
) => {
  let Comp: string;
  const name = (comp as any).name || (comp as any).displayName;

  if (comp === Canvas) {
    Comp = 'Canvas';
  }

  if (resolver[name]) {
    Comp = name;
  }

  for (let i = 0; i < Object.keys(resolver).length; i++) {
    const name = Object.keys(resolver)[i],
      fn = resolver[name];
    if (fn === comp) {
      Comp = name;
    }
  }

  if (typeof comp === 'string') {
    Comp = comp;
  }

  if (Comp === undefined) {
    throw new Error(ERROR_NOT_IN_RESOLVER.replace('%node_type%', name));
  } else {
    return Comp;
  }
};
```

image

ankri (approved):

ankri:
Thanks

Issue #273, opened 2021-07-08
Issue with drag and drop components between other components


hugominas:
**Describe the bug**
Its not possible to add new canvas items between other placed canvas items

**To Reproduce**

1. On the Landing Page example try placing containers after each other
2. then try and add a new container between them

**Expected behavior**
We should be able to place new components between other placed canvas components, this doesn't happen even when you add margin between them. This is crucial for a website builder, at the moment it either fails or places the new item on the top

**Screenshots**
![Craft js (1)](https://user-images.githubusercontent.com/3314986/124987045-4a09a780-e034-11eb-9d40-9750a8b6ce34.gif)

prevwong:
Interesting, it looks like because the Container is a Canvas node; when you try to drop something in between, it will be placed inside the Container:

<img width="1426" alt="Screenshot 2021-07-09 at 2 05 18 PM" src="https://user-images.githubusercontent.com/16416929/125031108-b8934780-e0be-11eb-8389-5911057c37fe.png">

Will look into it 👍

hugominas:
I tried with flex gap, margin and padding to solve this issues. @troelsfr in Discord posted a solution where he solved it by adding a new item to the DOM before and after the rendered node when the user is dragging. This has solved the issue. His implementation was done at the RenderNode and added 1.5rem height to the div

`  
const [dragging, setDragging] = useState(false);

  useEffect(() => {
    let onDragStart = () => setDragging(true);
    let onDragStop = () => setDragging(false);

    document.addEventListener("dragstart", onDragStart);
    document.addEventListener("dragend", onDragStop);
    return () => {
      document.removeEventListener("dragstart", onDragStart);
      document.removeEventListener("dragend", onDragStop);
    };
  });

 {dragging && <div className="dragSpacer"></div>}
      {render}
 {dragging && <div className="dragSpacer "></div>}
`

going to close this for now as it solved this issue.

![localhost_3003_matrizdesustentabilidade (2)](https://user-images.githubusercontent.com/3314986/125164384-d5488180-e189-11eb-8ff5-75b8829f15cd.gif)

hugominas:
I guess this issue is still present after all, it works and it seems to be related with this [issue](https://github.com/prevwong/craft.js/issues/274).

Everything works fine until you save the stated when you load, no matter where you drop your node it will always be stacked on the top of the tree. @prevwong any pointers on how we could solve this issue.

It looks like we are losing the dnd order of the nodes upon load.

[...]

PR #300, opened 2021-09-07
feat: improve dnd

prevwong:
Improves the dnd experience of the editor, particularly when trying to place an element between 2 other elements that have no margin between each other.

This PR also cleans up and fixes a couple of hidden issues with the event handling system (ie: duplicate event listeners being added to the same element). 

Related issue: #273 

Before:
https://loom.com/share/566e89ee0d6542f89f7f2d5f88461ff0

After:
https://www.loom.com/share/ae14a9dba36047fa8b88e3b0c4691e83


packages/core/src/editor/useInternalEditor.ts
```diff
@@ -26,22 +26,33 @@ export type useInternalEditorReturnType<C = null> = useCollectorReturnType<
...
+  useEffect(() => {
+    return () => {
+      if (!connectorInstance) {
+        return;
+      }
+
+      connectorInstance.cleanup();
+    };
+  }, [connectorInstance]);
+
```
prevwong, line 52: Cleanup any connectors used when the component unmounts. Previously, we only removed the connectors when the editor is enabled/disabled.

packages/core/src/editor/query.tsx
```diff
@@ -44,6 +44,8 @@ export function QueryMethods(state: EditorState) {
   return {
     /**
      * Determine the best possible location to drop the source Node relative to the target Node
+     *
+     * TODO: replace with Positioner.computeIndicator();
      */
     getDropPlaceholder: (
       source: NodeSelector,
```
prevwong, line 51: Can't replace this right now since `Positioner` requires the editor `store` as an input. Something we could look at once we merge the refactored editor store (#259) into this branch

matdru:
Like it, looks much better!

cypress/support/dnd.ts
```diff
@@ -199,14 +199,14 @@ Cypress.Commands.add(
         });
 
       if (position === 'inside') {
-        cy.get('@target').trigger('dragenter', {
+        cy.get('@target').trigger('dragover', {
```
timc1, line 202: Since this is going to fire on every tick of a drag event, is there any measurable way to check how this might affect the overall dnd performance?

prevwong: There's some optimisations in the `Positioner` class to prevent the indicator being updated on every tick, so that helps quite a bit. 

Not familiar with an automated way test this  - it's difficult to test dnd with Cypress because the best we can do there is emulate dnd events. We could just check the FPS count manually, and this new implementation is more or less the same as the previous one, so that should be good enough.

packages/core/src/editor/query.tsx
maziar-dandc: HMM :laughing:

Issue #225, opened 2021-03-21
Nested "type" key in Element props leads to serialization error


Jank1310:
The following props will cause a "Cannot convert undefined or null to object" error.

```
<SomeElement foo={{type:"bar"}} />
```
I suppose the following line causes it:
https://github.com/prevwong/craft.js/blob/2b09e3824cfb4f3b660a3f3bb14a14aba108ef34/packages/core/src/utils/serializeNode.tsx#L34

It the type key is nested one level deeper everything works as expected:
```
<SomeElement foo={{bar:{type:"bar"}}} />
```

eduludi:
I'm facing this issue too:

We integrating a rich text editor ([tiptap](https://tiptap.dev/)) as a Craf.js block. A Tiptap document has a `type` attribute in every node, causing this issue to raise after any attempt to change the document.

Another case was storing an image, where we had a `type` attribute for the mime type. We got it working by renaming it to `mime`. Any ideas on how to solve it or updates on this issue? 

And of course: Thanks for this awesome library!

iamtekeste:
@eduludi did you end up finding a fix for this? Facing the same issue while using tiptap editor.

ryanackley:
Why is this function even recursive? I couldn't find a test case that illustrates that point. Based on my understanding of the data model, it makes no sense to recurse on "props" which represent React node properties when serializing.

PR #308, opened 2021-09-12
Fix: Nested "type" key in Element props leads to serialization error

tscanlin:
This fix addresses the issue when `type` is nested as a key on the data that is used, resolving the following issues/PRs:
https://github.com/prevwong/craft.js/issues/225 and https://github.com/prevwong/craft.js/pull/251

### DIFF
--- a/packages/core/src/utils/serializeNode.tsx
+++ b/packages/core/src/utils/serializeNode.tsx
@@ -20,7 +20,7 @@ export const serializeComp = (
   props = Object.keys(props).reduce((result: Record<string, any>, key) => {
     const prop = props[key];
 
-    if (prop === undefined || prop === null) {
+    if (prop === undefined || prop === null || typeof prop === 'function') {
       return result;
     }
 
--- a/packages/core/src/utils/serializeNode.tsx
+++ b/packages/core/src/utils/serializeNode.tsx
@@ -31,7 +31,7 @@ export const serializeComp = (
         }
         return serializeComp(child, resolver);
       });
-    } else if (prop.type) {
+    } else if (typeof prop.type === 'function') {
       result[key] = serializeComp(prop, resolver);
     } else {
       result[key] = prop;


packages/core/src/utils/serializeNode.tsx
```diff
@@ -31,7 +31,7 @@ export const serializeComp = (
         }
         return serializeComp(child, resolver);
       });
-    } else if (prop.type) {
+    } else if (typeof prop.type === 'function') {
```
prevwong, line 34: Hm, this only works if the component is wrapped inside an object literal:
```tsx
<SomeComponent someObj={{ type: SomeOtherComponent }} />
```

But what about:
```tsx
<SomeComponent a={SomeOtherComponent} />
```

I actually can't really think of a strong use case where we need to serialise a component that's being passed as a prop. Perhaps, we should just ignore serialising any props that are not serialisable, including any components? So if we hit a `prop` that's a typeof function, we will just ignore it and move on to the next prop.

tscanlin: Yeah, skipping serializing any props that are functions makes sense to me. For this case though since the `type` prop is added internally by craft.js it seems like this PR should be enough to fix the issue, right?

Or what would the fix for handling any functions be?
I imagine, maybe just changing to something like this:

```
if (prop === undefined || prop === null || typeof prop === 'function') {
      return result;
}
```

prevwong: Yes, your solutions makes sense. 

I think we can skip serialising the `children` prop as well (since we're treating it the same way as functions right now). I feel like allowing this serialisation has always been troublesome.

https://github.com/prevwong/craft.js/blob/47336afd2c262312c093877409de249cb3683c67/packages/core/src/utils/serializeNode.tsx#L27-L35

Essentially, any prop that's not a JS literal (string, number, array, object etc) should be ignored. Ie:

```tsx
<SomeComponent 
  color="#333" // ok
  count={2} // ok
  values={{ someKey: 2, }} // ok
  someComp={AnotherComponent} // ignored
  someArray={[1,2,3]} // ok
  someComplexArray={[AnotherComponent]} // ignored
>
  Some Text // ok
  <AnotherComponent /> // ignored
</SomeComponent>
```

But perhaps this might need a bit more work to do. 

For now, we can just do the quick fix and ignore if the prop is a function.

prevwong (approved):
Thanks! 💯

tscanlin:
Thanks for merging this!

ryanackley:
This shows as being merged into prevwong/develop but when I look in github, the last commit to this file was in 2020. Someone may have force-pushed over this merge.

Issue #316, opened 2021-09-22
How do I save a node as a template that I can drag into the editor


linxianxi:
**Describe the solution you'd like**
save a node as a template that I can drag into the editor

Something like #209 needs to regenerate the ID of each node. But how do I save it in the sidebar and drag it?

The second parameter can now only be ReactElement, and it should be possible to customize the NodeTree type
```
connectors.create(
    ref,
    <Element />
  )
```
should like this
```
connectors.create(
    ref,
    // executed at the start of each drag to generate a new id for each node
   () => {
        const nodeTree = ...;
        const newNodeId = getRandomId();
        ......
        return nodeTree;
    }
  )
```

hugominas:
@linxianxi I have an open issue with this https://github.com/prevwong/craft.js/issues/311 I got stuck on the node serializations which I can't access from the available actions. Its only possible to serialize the state. prob @prevwong could point us in the correct direction.

linxianxi:
> @linxianxi I have an open issue with this #311 I got stuck on the node serializations which I can't access from the available actions. Its only possible to serialize the state. prob @prevwong could point us in the correct direction.

Maybe we can do it ourselves with these two methods in the source code?
![image](https://user-images.githubusercontent.com/47104575/135008095-7eb3d257-3845-4304-8436-e8ad50ea5238.png)

hugominas:
Yes, that's correct path. You need to clone the tree, so that you have new IDs, I did try serializing but it didn't work for me I guess it needs to be recursive to take care of all the child nodes and child of does child nodes. 

It would be great if we could reuse the internal method https://github.com/prevwong/craft.js/blob/3156043232b7e88f39e313b70f00bfb91ba830a3/packages/core/src/utils/serializeNode.tsx#L15 I will try rewriting the method for testing as soon as I have time

linxianxi:
@hugominas Of course, you need to traverse all nodes, childrens, and parents. If have the same ID, you need to generate a new ID to replace them

hugominas:
Almost got it working, https://github.com/prevwong/craft.js/issues/311#issuecomment-929317924

linxianxi:
@hugominas This example can save the template anywhere you want, but you can't drag it from the sidebar to the editor yet, maybe when my PR passes.

```
// import { getRandomId } from "@craftjs/utils";

const fromEntries = (pairs) => {
  if (Object.fromEntries) {
    return Object.fromEntries(pairs);
  }
[...]

PR #317, opened 2021-09-22
feat: support custom nodeTree in connectors.create()

linxianxi:
issue: #316 

To save nodes as templates that can be dragged into the editor, userElement can be of type function that returns NodeTree.

Executed at the start of each drag to generate a new id for each node.

```
connectors.create(
    ref,
    // executed at the start of each drag to generate a new id for each node
    () => {
        const nodeTree = ...;
        const newNodeId = getRandomId();
        ......
        return nodeTree;
    }
  )
```

### DIFF
--- a/packages/core/src/events/DefaultEventHandlers.ts
+++ b/packages/core/src/events/DefaultEventHandlers.ts
@@ -258,7 +258,7 @@ export class DefaultEventHandlers<O = {}> extends CoreEventHandlers<
             e.craft.stopPropagation();
             let tree;
             if (typeof userElement === 'function') {
-              let result = userElement();
+              const result = userElement();
               if (React.isValidElement(result)) {
                 tree = store.query.parseReactElement(result).toNodeTree();
               } else {
--- a/packages/core/src/events/DefaultEventHandlers.ts
+++ b/packages/core/src/events/DefaultEventHandlers.ts
@@ -258,12 +258,11 @@ export class DefaultEventHandlers<O = {}> extends CoreEventHandlers<
             e.craft.stopPropagation();
             let tree;
             if (typeof userElement === 'function') {
-              if (React.isValidElement(userElement())) {
-                tree = store.query
-                  .parseReactElement(userElement() as React.ReactElement)
-                  .toNodeTree();
+              let result = userElement();
+              if (React.isValidElement(result)) {
+                tree = store.query.parseReactElement(result).toNodeTree();
               } else {
-                tree = userElement();
+                tree = result;
               }
             } else {
               tree = store.query.parseReactElement(userElement).toNodeTree();
--- a/packages/core/src/events/DefaultEventHandlers.ts
+++ b/packages/core/src/events/DefaultEventHandlers.ts
@@ -1,6 +1,5 @@
 import { isChromium, isLinux } from '@craftjs/utils';
 import { isFunction } from 'lodash';
-import { isValidElement } from 'react';
 
 import { CoreEventHandlers, CreateHandlerOptions } from './CoreEventHandlers';
 import { Positioner } from './Positioner';
@@ -256,9 +255,10 @@ export class DefaultEventHandlers<O = {}> extends CoreEventHandlers<
           'dragstart',
           (e) => {
             e.craft.stopPropagation();
-            const tree = isValidElement(userElement)
-              ? store.query.parseReactElement(userElement).toNodeTree()
-              : userElement();
+            const tree =
+              typeof userElement === 'function'
+                ? userElement()
+                : store.query.parseReactElement(userElement).toNodeTree();
 
             const dom = e.currentTarget as HTMLElement;
             this.draggedElementShadow = createShadow(
--- a/packages/core/src/events/DefaultEventHandlers.ts
+++ b/packages/core/src/events/DefaultEventHandlers.ts
@@ -1,5 +1,6 @@
 import { isChromium, isLinux } from '@craftjs/utils';
 import { isFunction } from 'lodash';
+import React from 'react';
 
 import { CoreEventHandlers, CreateHandlerOptions } from './CoreEventHandlers';
 import { Positioner } from './Positioner';
@@ -245,7 +246,7 @@ export class DefaultEventHandlers<O = {}> extends CoreEventHandlers<
       },
       create: (
         el: HTMLElement,
-        userElement: React.ReactElement | (() => NodeTree),
+        userElement: React.ReactElement | (() => NodeTree | React.ReactElement),
         options?: Partial<CreateHandlerOptions>
       ) => {
         el.setAttribute('draggable', 'true');
--- a/packages/core/src/events/DefaultEventHandlers.ts
+++ b/packages/core/src/events/DefaultEventHandlers.ts
@@ -1,6 +1,6 @@
 import { isChromium, isLinux } from '@craftjs/utils';
 import { isFunction } from 'lodash';
-import React from 'react';
+import React, { isValidElement } from 'react';
 
 import { CoreEventHandlers, CreateHandlerOptions } from './CoreEventHandlers';
 import { Positioner } from './Positioner';
@@ -256,10 +256,18 @@ export class DefaultEventHandlers<O = {}> extends CoreEventHandlers<
           'dragstart',
           (e) => {
             e.craft.stopPropagation();
-            const tree =
-              typeof userElement === 'function'
-                ? userElement()
-                : store.query.parseReactElement(userElement).toNodeTree();
+            let tree;
+            if (typeof userElement === 'function') {
+              if (isValidElement(userElement())) {
+                tree = store.query
+                  .parseReactElement(userElement() as React.ReactElement)
+                  .toNodeTree();
+              } else {
+                tree = userElement();
+              }
+            } else {
+              tree = store.query.parseReactElement(userElement).toNodeTree();
+            }
 
             const dom = e.currentTarget as HTMLElement;
             this.draggedElementShadow = createShadow(
--- a/packages/core/src/events/DefaultEventHandlers.ts
+++ b/packages/core/src/events/DefaultEventHandlers.ts
@@ -1,6 +1,6 @@
 import { isChromium, isLinux } from '@craftjs/utils';
 import { isFunction } from 'lodash';
-import React, { isValidElement } from 'react';
+import React from 'react';
 
 import { CoreEventHandlers, CreateHandlerOptions } from './CoreEventHandlers';
 import { Positioner } from './Positioner';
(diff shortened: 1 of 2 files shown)


packages/core/src/events/DefaultEventHandlers.ts
```diff
@@ -245,7 +245,7 @@ export class DefaultEventHandlers<O = {}> extends CoreEventHandlers<
       },
       create: (
         el: HTMLElement,
-        userElement: React.ReactElement,
+        userElement: React.ReactElement | (() => NodeTree),
```
prevwong, line 248: Could we have `userElement: React.ReactElement | NodeTree` instead?

In that case, we will probably need a helper function (ie: `isNodeTree`) to check if the input is a valid `NodeTree` as well.

linxianxi: You cannot replace it with NodeTree.Because this function needs to be executed each time to generate a new ID for each node.If it is NodeTree, the id of each drag will be the same

Spenc84:
What is the status of this PR? This change would be incredibly helpful for our use case so we would love to see it go through. Is there anything that still needs to happen before this can get merged in?

prevwong:
Apologies on the late response. Was out on holiday! PR looks good, would be great to have [this](https://github.com/prevwong/craft.js/pull/317#discussion_r781045160) included as well if possible!

linxianxi:
> Apologies on the late response. Was out on holiday! PR looks good, would be great to have [this](https://github.com/prevwong/craft.js/pull/317#discussion_r781045160) included as well if possible!

It's ok now.

Spenc84:
@linxianxi Would you be willing to slightly adjust your most recent commit so that userElement() is only getting called once? That should improve efficiency slightly and could also prevent a potential issue that might pop up in the unlikely scenario that userElement() returns a React Element on one call and a Node Tree on the other.

Something like this perhaps?
```
let tree, result;
if (typeof userElement === 'function') {
    result = userElement();
    if (isValidElement(result)) {
        tree = store.query
            .parseReactElement(result as React.ReactElement)
            .toNodeTree();
    } else {
        tree = result;
    }
} else {
    tree = store.query.parseReactElement(userElement).toNodeTree();
}

**Comment by linxianxi** [comment]
> @linxianxi Would you be willing to slightly adjust your most recent commit so that userElement() is only getting called once? That should improve efficiency slightly and could also prevent a potential issue that might pop up in the unlikely scenario that userElement() returns a React Element on one call and a Node Tree on the other.
> 

yes,I wrote it too fast to notice

**Review by prevwong — APPROVED** [review]
Thanks! 💯  @linxianxi and @Spenc84

**Comment by prevwong** [comment]
Released in 0.2.0-beta.2 🎉

**Comment by linxianxi** [comment]
> Released in 0.2.0-beta.2 🎉

can use it in next.js branch?

**Comment by Nym02** [comment]
@linxianxi Do I need to put anything into the resolver? Is there any code example of this?

**Comment by Nym02** [comment]
@linxianxi  @Spenc84 @prevwong 

_**"{\"ROOT\":{\"type\":{\"resolvedName\":\"CanvasContainer\"},\"isCanvas\":true,\"props\":{\"flexDirection\":\"column\",\"alignItems\":\"flex-start\",\"justifyContent\":\"flex-start\",\"fillSpace\":\"no\",\"padding\":[\"0\",\"20\",\"0\",\"20\"],\"margin\":[\"0\",\"0\",\"0\",\"0\"],\"background\":\"#FFFFFF\",\"color\":\"#000000\",\"shadow\":0,\"radius\":0,\"width\":\"600px\",\"minHeight\":\"300px\"},\"displayName\":\"Frame Container\",\"custom\":{\"displayName\":\"App\"},\"hidden\":false,\"nodes\":[],\"linkedNodes\":{}}}"**_ 

this is a JSON i get when I save my template. Now I want to drag this to the editor? Is it possible? I can see that this PR supports NodeTree in **connectors.create()** . Is there anyway I can implement drag feature on Serialized JSON?

**Comment by xs-nayeem** [comment]
> @linxianxi Would you be willing to slightly adjust your most recent commit so that userElement() is only getting called once? That should improve efficiency slightly and could also prevent a potential issue that might pop up in the unlikely scenario that userElement() returns a React Element on one call and a Node Tree on the other.
> 
> Something like this perhaps?
> 
> ```
> let tree, result;
> if (typeof userElement === 'function') {
>     result = userElement();
>     if (isValidElement(result)) {
>         tree = store.query
>             .parseReactElement(result as React.ReactElement)
>             .toNodeTree();
>     } else {
>         tree = result;
>     }
> } else {
>     tree = store.query.parseReactElement(userElement).toNodeTree();
> }
> ```

@Spenc84 It looks like it only takes a single node. what to do in case of multiple node? Are we going to see any fix for this?

**Comment by Spenc84** [comment]
@xs-nayeem  I apologize, it's been a couple of years since I worked on the section of our codebase that implements craft, so my memory of the implementation details is a bit hazy, but I thought any single node could specify any number of of child nodes underneath it? If so, it seems like you simply need to make sure that the single node used here is the parent of any of the other nodes that should also get added to the layout. Just make sure that your new node and every one of it's descendants each have their own unique id.

**Comment by xs-nayeem** [comment]
> @xs-nayeem I apologize, it's been a couple of years since I worked on the section of our codebase that implements craft, so my memory of the implementation details is a bit hazy, but I thought any single node could specify any number of of child nodes underneath it? If so, it seems like you simply need to make sure that the single node used here is the parent of any of the other nodes that should also get added to the layout. Just make sure that your new node and every one of it's descendants each have their own unique id.

@Spenc84 thanks for your response.  Well I am trying to achieve that but somehow I am missing something maybe.  
[https://pastebin.com/gzqGpyu5](https://pastebin.com/gzqGpyu5) this is the JSON i am trying to drag to the editor. If could make some time to give me some hint about how can i drag this to the editor.

**Comment by Spenc84** [comment]
Again I haven't used any of that code in awhile, so my memory is hazy, but if that JSON you linked is static and you're trying to replicate that after each drag then I imagine you'll run into some issues with duplicated ids. IIRC each node in your layout has to have it's own unique ID, so for instance you couldn't have two different nodes with the id "ROOT", or two different nodes with the id "b2Hu6-fsIl" within the same layout JSON. If you're not doing it already, my first suggestion would be to write some code that parses that JSON and replaces each id with a new unique id within the callback method that you're passing to .create(ref, callback).

Example:
``` js
connectors.create(
    ref,
    // executed at the start of each drag to generate a new id for each node
    () => {
        /* Here you might copy the template JSON you're trying to replicate while also
        *  replacing the id for every node with a new unique one. */
        const newNodeTree = generateUniqueNodeTree( EXISTING_STATIC_JSON );
        return newNodeTree ;
    }
  )
```

I also don't recall exactly how the code was implemented to integrate the JSON for a new nodeTree into the existing layout JSON via the .create method. I imagine it would take all of the nodes in the returned nodeTree JSON and add it to the existing layout JSON, and then take the id of the first node in returned JSON and add that id as a child of whichever node on the layout that you dragged the new nodeTree into.

However if it does not work that way then I suppose that it's also entirely possible that you can only return a single node from that callback, in which case you'd just let the create method handle adding the root node of your new nodeTree to the existing JSON and then you'd have to add all of the descendants of that node yourself programmatically either during or after the drag event which (IIRC) shouldn't be to difficult to do. You will still need to ensure that each newly added node has it's own unique id.

thomassmartinez:
I'd like to know if mutating JSON is advised. I ask because I've implemented a feature to copy and paste fields from other JSON objects. However, I always generate new ids and assign them to both the parents and nodes. However, when I'm dragging a field from within a pasted group, it duplicates the node. Essentially, it doesn't remove the node from the nodes.

before 
```
{
   "ROOT": {
      "type": {
         "resolvedName": "Container"
      },
      "isCanvas": true,
      "props": {},
      "displayName": "Container",
      "custom": {},
      "hidden": false,
      "nodes": [
         "KRI1PuP_ff"
      ],
      "linkedNodes": {}
   },
   "xkui96ghtpc": {
      "type": {
         "resolvedName": "Group"
      },
      "isCanvas": true,
      "props": {
         "name": "Agrupamento",
         "label": "parent group",
         "visible": true
      },
      "displayName": "Group",
      "custom": {},
      "parent": "KRI1PuP_ff",
      "hidden": false,
      "nodes": [
         "48pgpnx7q88"
      ],
      "linkedNodes": {}
   },
   "48pgpnx7q88": {
      "type": {
         "resolvedName": "TextField"
      },
      "isCanvas": false,
      "props": {
         "label": "text parent",
         "placeholder": "Digite aqui...",
         "type": "text",
         "name": "Texto curto",
         "visible": true
      },
      "displayName": "TextField",
      "custom": {},
      "parent": "KRI1PuP_ff",
      "hidden": false,
      "nodes": [],
      "linkedNodes": {}
   },
   "KRI1PuP_ff": {
      "type": {
         "resolvedName": "Group"
      },
      "isCanvas": true,
      "props": {
         "name": "Agrupamento",
         "label": "anyway",
         "visible": true
      },
      "displayName": "Group",
      "custom": {},
      "parent": "ROOT",
      "hidden": false,
      "nodes": [
         "xkui96ghtpc"
      ],
      "linkedNodes": {}
   }
}
```

"After dragging the field."

```
{
   "ROOT": {
      "type": {
         "resolvedName": "Container"
      },
      "isCanvas": true,
      "props": {},
      "displayName": "Container",
      "custom": {},
      "hidden": false,
      "nodes": [
         "KRI1PuP_ff"
      ],
      "linkedNodes": {}
   },
   "xkui96ghtpc": {
      "type": {
         "resolvedName": "Group"
      },
      "isCanvas": true,
      "props": {
         "name": "Agrupamento",
         "label": "parent group",
         "visible": true
      },
      "displayName": "Group",
      "custom": {},
      "parent": "KRI1PuP_ff",
      "hidden": false,
      "nodes": [
         "48pgpnx7q88"
      ],
      "linkedNodes": {}
   },
   "48pgpnx7q88": {
      "type": {
         "resolvedName": "TextField"
      },
      "isCanvas": false,
      "props": {
         "label": "text parent",
         "placeholder": "Digite aqui...",
         "type": "text",
         "name": "Texto curto",
         "visible": true
      },
      "displayName": "TextField",
      "custom": {},
      "parent": "KRI1PuP_ff",
      "hidden": false,
      "nodes": [],
      "linkedNodes": {}
   },
   "KRI1PuP_ff": {
      "type": {
         "resolvedName": "Group"
      },
      "isCanvas": true,
      "props": {
         "name": "Agrupamento",
         "label": "anyway",
         "visible": true
      },
      "displayName": "Group",
      "custom": {},
      "parent": "ROOT",
      "hidden": false,
      "nodes": [
         "48pgpnx7q88",
         "xkui96ghtpc" <--- the issue is here.
      ],
      "linkedNodes": {}
   }
}
```

Notice that it didn't remove from the node after dragging.

image

Issue #318, opened 2021-09-22
I need a callback to modify the property before or after the element is moved


linxianxi:
**Is your feature request related to a problem? Please describe.**
I have a Column component that needs to change its width before or after I move it.
Move the second column of the top row to the middle of the bottom row

![image](https://user-images.githubusercontent.com/47104575/134443029-160cf9cb-5409-470a-8cc6-a703a9f37bdd.png)

api like this
```
onBoforeMoveEnd(targetNode, currentNode) 
```

PR #320, opened 2021-09-22
feat: add onBeforeMoveEnd

linxianxi:
issue: #318

### DIFF
--- a/packages/core/src/editor/Editor.tsx
+++ b/packages/core/src/editor/Editor.tsx
@@ -41,7 +41,7 @@ export const Editor: React.FC<Partial<Options>> = ({
       },
       (value) => value !== undefined
     );
-  }, [enabled, indicator, onNodesChange, onRender, resolver]);
+  }, [enabled, indicator, onBeforeMoveEnd, onNodesChange, onRender, resolver]);
 
   const context = useEditorStore(
     options,
--- a/packages/core/src/editor/Editor.tsx
+++ b/packages/core/src/editor/Editor.tsx
@@ -16,6 +16,7 @@ export const Editor: React.FC<Partial<Options>> = ({
   children,
   onRender,
   onNodesChange,
+  onBeforeMoveEnd,
   resolver,
   enabled,
   indicator,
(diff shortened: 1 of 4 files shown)


packages/core/src/interfaces/editor.ts
```diff
@@ -10,6 +10,7 @@ import { CoreEventHandlers } from '../events';
 
 export type Options = {
   onRender: React.ComponentType<{ render: React.ReactElement }>;
+  onBeforeMoveEnd: (targetNode: Node, currentNode: Node) => void;
```
prevwong, line 13: Based on the function call in the `actions` file above, this should be `onBeforeMoveEnd: (newParentNode: Node, targetNode: Node)`, right? 

Perhaps we should also include the existingParentNode as well? So something like: `onBeforeMoveEnd: (targetNode: Node, newParentNode: Node, existingParentNode: Node)`. This way related to your issue #318, when a user moves a Node, you can update the width of the Node that was moved, and you can also update the widths of the children of the existing and new parent node.

prevwong (commented):
Thanks for this PR! Think this option could be quite handy. Could you also add some tests here: https://github.com/prevwong/craft.js/blob/next/packages/core/src/editor/tests/actions.test.ts#L447

packages/core/src/editor/Editor.tsx
```diff
@@ -30,7 +31,14 @@ export const Editor: React.FC<Partial<Options>> = ({
...
+        onRender,
+        onNodesChange,
+        onBeforeMoveEnd,
+        resolver,
+        enabled,
+        indicator,
+      },
       (value) => value !== undefined
     );
   }, [enabled, indicator, onNodesChange, onRender, resolver]);
```
ankri, line 44: Does `onBeforeMoveEnd` need to be a part of the `useMemo` dependency array?

ankri (commented):
I like @prevwong's suggestion. Also please check if `onBeforeMoveEnd` needs to be part of the `useMemo` dependency array.

linxianxi:
@prevwong  Please review again.

ankri (approved):

prevwong (approved):
LGTM 💯  Thanks!

PR #327, opened 2021-10-05
Fix the paths and command

hariombalhara:
`npm dev` command doesn't exist now
`basic` folder location has changed. So fix it according to that.


ankri (approved):
lgtm, thanks

prevwong (approved):
Thanks!

PR #330, opened 2021-10-08
fix: add generic props type for NodeData

kkshyu:
in typescript, we suppose not to use `any`. we can simply add the generic type on Node to make codes better.
ex.
```ts
  const {
    actions: { setProp },
    props,
  } = useNode((node: Node<CraftProgramCollectionProps>) => ({
    props: node.data.props,
  }))
```
`setProps` will be:
```
const setProp: (cb: (props: CraftProgramCollectionProps) => void, throttleRate?: number | undefined) => void
```


Nedi11:
+1

Issue #229, opened 2021-04-09
Impossible to stop the creation of a component dragged from the toolbox


dentiluca:
**Is your feature request related to a problem? Please describe.**
If I start dragging a component from the toolbox, I can't interrupt the creation of the component in the canvas.

**Describe the solution you'd like**
For example, I could press ESC and the dragged component won't be dropped in the canvas, or I could move the mouse out of the canvas.

ggunti:
I remarked same thing. This happens also in the example at https://craft.js.org.
I made a screen recording where is visible that I try to move the cursor out of the canvas, but the component is still created:


https://user-images.githubusercontent.com/26307699/114759315-aecfa180-9d66-11eb-83f6-22fb17cd7882.mov

Shaddix:
Faced this issue as well.
If there are any suggestions on how it could be implemented/workarounded from outside of craft.js, I'd be glad to hear :)

kenkz447:
+1

ksouthworth:
+1

wesAlves:
+1

th-manas:
@prevwong Do we have any updates on this issue please? How can we solve this issue any workaround or suggestion.

PR #344, opened 2021-11-15
Fix: Stop the creation of a component when mouse leaves Frame

microfe:
Issue: #229

### DIFF
--- a/packages/core/src/events/DefaultEventHandlers.ts
+++ b/packages/core/src/events/DefaultEventHandlers.ts
@@ -160,7 +160,18 @@ export class DefaultEventHandlers<O = {}> extends CoreEventHandlers<
             store.actions.setIndicator(indicator);
           }
         );
+        
+        const unbindDragLeave = this.addCraftEventListener(
+          el,
+          'dragleave',
+          (e) => {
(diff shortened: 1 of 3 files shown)


packages/core/src/events/DefaultEventHandlers.ts
```diff
@@ -160,7 +160,18 @@ export class DefaultEventHandlers<O = {}> extends CoreEventHandlers<
...
         );
+        
+        const unbindDragLeave = this.addCraftEventListener(
+          el,
+          'dragleave',
+          (e) => {
+            e.craft.stopPropagation();
+            e.preventDefault();
 
+            this.positioner?.removeIndicator();
```
prevwong, line 171: Could you remove optional chaining for now? (We're still on an older version of TS) 

```ts
if ( !this.positioner ) {
  return; 
}

this.positioner.removeIndicator();
```

Do you mind explaining to me a little on how this works? It seems we're removing the indicator when the dragged element leaves a droppable element. Let's say if we're dragging some elements within the ROOT element, then we move our cursor out of the ROOT element, would this remove the indicator as well?

PR #360, opened 2022-01-10
fix(0.1.x): set drag state onDragStart

### LINKED ISSUE #321: EditorState "dragged" event not being set
**Describe the bug**
When a node is being dragged, the EditorState `dragged` event is not being set to NodeId - it always remains `null`.

**To Reproduce**
Adding this to the `useEditor` hook - the `dragged` always remains `null`. Tested this within the Basic example from the craftjs repo (within components/Toolbox.js). The other two events are being set correctly.

```js
const { connectors } = useEditor((state) => {
    console.log("Dragged: ", state.events.dragged);
    console.log("Selected: ", state.events.selected);
    console.log("Hovered: ", state.events.hovered);
 });
```

**Expected behavior**
I expect the `dragged` event to be correctly set to the dragged node's NodeId within `useEditor`.

### Your environment

| craft.js         | 0.1.0-beta.20
| React            | 17.0.2
| Browser          | Chrome, Firefox
| Operating System | MacOS

prevwong:
Closes #321

### DIFF
--- a/packages/core/src/events/DefaultEventHandlers.ts
+++ b/packages/core/src/events/DefaultEventHandlers.ts
@@ -119,7 +119,9 @@ export class DefaultEventHandlers extends CoreEventHandlers {
           (e) => {
             e.craft.stopPropagation();
 
-            const { query } = store;
+            const { query, actions } = store;
+
+            actions.setNodeEvent('dragged', id);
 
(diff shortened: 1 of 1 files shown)


ankri:
Simple, yet effective 👍  :)

PR #372, opened 2022-02-07
fix: remove redundant "}" of the image url

clyer:
now, all images in [documentation](https://craft.js.org/docs/guides/basic-tutorial) are broken by redundant `%7D` in the end.

example:
- https://craft.js.org/img/tutorial/interface.png%7D


prevwong (approved):
Thanks!

PR #374, opened 2022-02-09
feat: add events.draggedOver

### LINKED ISSUE #28: Add draggedOver to state.events

### LINKED ISSUE #183: Is there a way to detect when a child component is incoming into a parent container?

ankri:
Hi there,

PR #186 is long overdue but I couldn't get it to run because of some stupid git merge conflicts.  So here we go again.

Here is the old text:
This PR adds draggedOver to the events. It's almost the same as #37 but targets the next branch and the new EventHandlers, which also gives us a cleaner API. (#37 will be obsolete - #186 will also be obsolete ).

Since the new EventHandlers introduced a Set as the store for the EventHandlers, we can now store not only the id of the node we are dragging over, but all of its ancestors, too.

Closes #28
Closes #183

Supersedes #186
Supersedes #37

### DIFF
--- a/examples/basic/components/user/Card.js
+++ b/examples/basic/components/user/Card.js
@@ -1,4 +1,4 @@
-import { Element, useNode } from '@craftjs/core';
+import { Element, useNode, useEditor } from '@craftjs/core';
 import React from 'react';
 
 import { Button } from './Button';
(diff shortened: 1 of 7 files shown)


vazitron:
> 

I see your point.
On your screenshot, the mousepointer (albeit not visible) is hovering over the <Card /> element, which results in it being higlighted as draggedOver.
However, because the drop indicator is positioned below the <Card />, the Card should not be probably highlighted.

I personally would expect the draggedOver behave as "Is drop indicator inside me (or inside some of my children)".
If that passes, we should also rename it to something like "containsDropIndicator" or "isDropTarget", which is more descriptive of what actually happens, and it also avoids confusion with how the native dragOver event behaves.

noahbaron91:
Any updates on this?

mattvb91:
Would love to have this too!

geohuz:
Any progress to use the draggedOver?

ankri:
Closing due to no time available

PR #454, opened 2022-08-30
Updated `isMultiSelectEnabled` to support ctrl key on windows

amkoehler:
Hello - while testing beta, I found that the use of `metaKey` for multi-selection does not work well on Windows. Generally the commands for metaKey on mac are equivalent to the ctrl key on windows, so I added a platform check using `navigator.userAgent`.


asyura:
ctrl key on windows still not working

PR #470, opened 2022-11-22
chore: support tree shaking for lodash

trionfordring:
Hello~
replace `import { xxx } from 'lodash';` to `import xxx from 'lodash/xxx';`


prevwong:
Thank you! :100 

Duplicate of #540

PR #509, opened 2023-04-07
hotfix: optional enabled prop

### LINKED ISSUE #508: Enabled prop is required since last release
If you updated to the latest version of craft, you might see that the editor is disabled. Thats because the prop "enabled" of Editor is required now. You have to explicitly tell the editor that you want to enable it. 

I'm not sure if this is by design, but this https://github.com/prevwong/craft.js/commit/c74f4c94e1722c4f515a34ebf8718af896661de7 
affected it.

@prevwong lmk if this is intentionally so I can update the docs, or if its an issue  so I can see and fix this.

prevwong:
Fix regression caused in #501 where `enabled` option is incorrectly set to undefined if the option was not explicitly specified in the `<Editor />` component

Closes #508

### DIFF
--- a/packages/core/src/editor/Editor.tsx
+++ b/packages/core/src/editor/Editor.tsx
@@ -72,11 +72,14 @@ export const Editor: React.FC<React.PropsWithChildren<Partial<Options>>> = ({
 
   // sync enabled prop with editor store options
   useEffect(() => {
-    if (!context) {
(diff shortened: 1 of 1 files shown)


ismajl-ramadani (approved):
Tested it locally, and it's working now, meaning that enabled is "true" default and its optional.

Issue #527, opened 2023-05-14
Reduce bundle size: Use individual packages like `lodash.identity` instead of complete `lodash`


chavda-bhavik:
**Is your feature request related to a problem? Please describe.**
It's hard to see that while rendering a page with craftjs, 1/4 portion contains **lodash** with a `531KB` size.
![image](https://github.com/prevwong/craft.js/assets/50201755/77371426-f08a-46ae-a772-8bbabc4416ce)

On the other side, we're using only a few facilities of `lodash`. 
https://github.com/search?q=repo%3Aprevwong%2Fcraft.js%20lodash&type=code

**Describe the solution you'd like**
Instead of using `lodash` we can use individual packages, like,
```
// instead of using
import flatten from 'lodash/flatten';

// use of this will keep bundle size low
import flatten from 'lodash.flatten'
```

**Additional context**
All the used lodash utilities are available as individual packages. Here is the complete list of utilities needed,
1. `lodash.flatten`
2. `lodash.forin`
3. `lodash.clonedeep`
4. `lodash.clonedeep`
5. `lodash.pickby`
6. `lodash.isequalwith`
7. `lodash.isfunction`
8. `lodash.identity`
9. `lodash.mapvalues`

ismajl-ramadani:
Hey @chavda-bhavik, thanks for opening this issue. There are some issues with the related content already posted to the Craft.js repo, but it seems that they're a little bit outdated.
https://github.com/prevwong/craft.js/pull/470
https://github.com/prevwong/craft.js/pull/256

Would you like to improve, all of these imports and create a PR?

Meanwhile, I'm closing this issue since it's not related to the core functionality of Craft.js.

PR #540, opened 2023-06-17
chore: import specific lodash methods

chavda-bhavik:
# Issue
https://github.com/prevwong/craft.js/issues/527

*Note: To make a successful build I have to update https://github.com/prevwong/craft.js/blob/0bc86651f6ffb2855edeba9aa300645007790d23/packages/layers/package.json#L43 version to `^0.2.0-beta.11`. Which I have not added in the commit. Please let me know if it's needed.


prevwong (commented):
Thanks for the PR! 

It looks like this is discouraged by lodash: https://lodash.com/per-method-packages

How about doing one of the following:
1. Import only the relevant lodash file:
```tsx
import mapValues from 'lodash/mapValues
```
I know there are some lodash functions in the codebase not being imported like this, so perhaps we can try updating those and see the bundle size?

2. Alternatively, we can try the `lodash-es` package.

 Perhaps it might be worth trying both approaches and comparing the bundle size in the end.

chavda-bhavik:
Hey @prevwong the `lodash-es` import causes to fail the build so I've updated imports with `lodash/*`.

prevwong (approved):
Thank you! 💯

PR #543, opened 2023-06-20
Resolve issue with indicator width:

brendanowens:
Resolves issue where the indicator would be a negative width if the `Layers` plugin was not placed on the left-hand side of the screen. This was due to `getBoundingClientRect()` calculating the `headingPos` realtive to the viewport. This change accounts for the `Layers` component placed anywhere in the viewport by adding back in the postion of the layer parent item.

See screenshot for real-world implementation that caused the indicator to the inccorect width due to the `Layers` component being used elsewhere in the viewport:
Screenshot 2023-06-20 at 4 05 45 PM


prevwong (approved):
Thank you! 💯

Issue #545, opened 2023-07-04
Customizing `DefaultEventHandlers` in editor is not working


chavda-bhavik:
**Describe the bug**
`Editor` allows providing `DefaultEventHandlers` that is useful for cases like, `Want to wrap elements into section when any element is dropped in ROOT`.

**To Reproduce**
1. In `Editor` provide `DefaultEventHandlers` like,
```
<Editor
      handlers={(store) =>
        new DefaultEventHandlers({
          // @ts-ignore
          store,
          isMultiSelectEnabled: (e: MouseEvent) => !!e.metaKey,
        })
      }
>
...
</Editor>
```
Here the `DefaultEventHandlers` is written manually to handle the needs.

**Expected behavior**
Provided `DefaultEventHandlers` should be used.

**Additional context**
I found out the issue is in [useEditorState](https://github.com/prevwong/craft.js/blob/6e3647ed9e7f27f1cff82b6b3d8d81d2b6d94db3/packages/core/src/editor/store.tsx#L103), where we need to provide `handlers` like,
```
export const useEditorStore = (
  options: Partial<Options>,
  patchListener: PatchListener<
    EditorState,
    typeof ActionMethodsWithConfig,
    typeof QueryMethods
  >
): EditorStore => {
  // TODO: fix type
  return useMethods(
    ActionMethodsWithConfig,
    {
      ...editorInitialState,
      ...(options.handlers ? { handlers: options.handlers } : {}),
      options: {
        ...editorInitialState.options,
        ...options,
      },
    },
    QueryMethods,
    patchListener
  ) as EditorStore;
};
```

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| craft.js         |  `0.2.0-beta.8`
| React            | `18.2.0`
| TypeScript       | `^4.9.5`
| Browser          | `Chrome`
| npm/Yarn         | `yarn 1.22.19`
| Operating System | `Linux`

prevwong:
Isn't the overwritten `DefaultEventHandlers` that you passed to the `Editor` component used? I just tried it out on the basic example, and it does seem to work. Unless I am missing something?

PR #546, opened 2023-07-11
`fix:` custom handlers not accepting

chavda-bhavik:
- Fix the issue of providing custom `DefaultEventHandlers` not working.

https://github.com/prevwong/craft.js/issues/545

### DIFF
--- a/packages/core/src/editor/store.tsx
+++ b/packages/core/src/editor/store.tsx
@@ -101,6 +101,7 @@ export const useEditorStore = (
     ActionMethodsWithConfig,
     {
       ...editorInitialState,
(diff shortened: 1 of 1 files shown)


packages/core/src/editor/store.tsx
```diff
@@ -101,6 +101,7 @@ export const useEditorStore = (
     ActionMethodsWithConfig,
     {
       ...editorInitialState,
+      ...(options.handlers ? { handlers: options.handlers } : {}),
```
prevwong, line 104: I'm pretty sure this doesn't do anything since the `handlers` property is no longer part of the state

Issue #549, opened 2023-07-24
Duplicate Node Ids inside serialised nodes state


AminRafaey:
**Describe the bug**
When using `query.parseReactElement(<Element id={getUniqueId()} is={ContainerComponent} canvas></Element>).toNodeTree()`.  Rarely when users complain after observing their craft state we realised they end up with duplicate node ids. 

**To Reproduce**
This is an infrequent use case so there are no specific steps I could mention which end up with an error.

**Expected behavior**
A function craft.js is used for generating a unique id, it should always generate a unique id.

**Additional context**
After looking inside the source code we believe this function could be a reason:

`import { nanoid } from 'nanoid';

// By default nanoid generate an ID with 21 characters. To reduce the footprint, we default to 10 characters.
// We have a higher probability for collisions, though

/**
 * Generate a random ID. That ID can for example be used as a node ID.
 *
 * @param size The number of characters that are generated for the ID. Defaults to `10`
 * @returns A random id
**/

export const getRandomId = (size: number = 10) => nanoid(size);`

Moreover, we don't have any control over our app to make it unique on our own because craft.js doesn't respect if we provide a unique id, It uses its own generated unique id. You can see an example here

`export function createNode(
  newNode: FreshNode,
  normalize?: (node: Node) => void
) {
  let actualType = newNode.data.type as any;
  let id = newNode.id || getRandomNodeId();`

It looks like it is respecting our id, but if you see here:

`return createNode(
    {
      data: {
        type: actualType,
        props: { ...element.props },
      },
    },
    (node) => {
      if (normalize) {
        normalize(node, element);
      }
    }
  );`

It's not passing an id forward. @prevwong is it on purpose or is it a bug?

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| craft.js         | v0.2.0-beta.0
| React            | ^16.13.1
| TypeScript       | ^4.7.2
| npm/Yarn         | npm

Lublu1:
+1

PR #553, opened 2023-08-21
fix: Custom id passing forward from Parse Node From JSX

AminRafaey:
- Fix: Custom id passing forward from Parse Node From JSX
[549](https://github.com/prevwong/craft.js/issues/549)

### DIFF
--- a/packages/core/src/utils/parseNodeFromJSX.tsx
+++ b/packages/core/src/utils/parseNodeFromJSX.tsx
@@ -18,6 +18,7 @@ export function parseNodeFromJSX(
 
   return createNode(
     {
+      id: element.props?.id,
       data: {
         type: actualType,
         props: { ...element.props },


prevwong:
Since we're dealing with the `<Element />` specifically, we could add `id: null` here:

https://github.com/prevwong/craft.js/blob/6106115329d88bfa592012e810356c32f7de5b6d/packages/core/src/nodes/Element.tsx#L11-L16

Then the logic in `createNode` should handle the rest:
https://github.com/prevwong/craft.js/blob/6106115329d88bfa592012e810356c32f7de5b6d/packages/core/src/utils/createNode.ts#L63-L70

AminRafaey:
@prevwong It's not gonna work like this. Because in this code `if` condition is gonna run and it will put `id` inside `node.data`.
https://github.com/prevwong/craft.js/blob/6106115329d88bfa592012e810356c32f7de5b6d/packages/core/src/utils/createNode.ts#L63-L70
But we need it inside a node, Here is the node interface.
https://github.com/prevwong/craft.js/blob/6106115329d88bfa592012e810356c32f7de5b6d/packages/core/src/interfaces/nodes.ts#L26C1-L34C3

prevwong:
The `id` prop in `<Element id="" />` is currently being used to set the id for a linked node, which is not the same as `node.id`; which puts us in a bit of a difficult position as this would further cause confusion.

The desired results of this PR could also be achieved like so:

```tsx
query
.parseReactElement(<Element id={getUniqueId()} is={ContainerComponent} canvas></Element>)
.toNodeTree(node => {
   node.id = "some custom id";
});
```

fakob:
Thank you @prevwong your solution saved my day!

PR #563, opened 2023-09-26
(layers) Fix for dropping nodes into empty container

### LINKED ISSUE #562: Unable to drop node into container in layers panel when container has 0 children
**Describe the bug**
A new container that is placed on the canvas with 0 child nodes is unable to accent child nodes using drag-and-drop within the Layers package.

**To Reproduce**

1. Run the example landing page locally
2. Drop a new container onto the canvas
3. Drop another node (I.e. text block) onto the canvas next to the container
4. Attempt to drop the node from step (3) into the container using the layers panel on the right side
5. The node cannot be placed within the container using the layers panel

**Expected behavior**
A node should be able to be dropped into a container using the layers panel, even if the container has no other children.

**Screenshots**
[Recording demo of issue](https://www.loom.com/share/e5eb7aec08484eb3b8a19898664ff178?sid=14ceaee9-a034-4837-bc79-31b50d8153b1)

brendanowens:
Fixes #562

This PR resolves an issue in the layers package where a node was not able to be placed within a container using the layers drag and drop interface when the accepting container had no children

### DIFF
--- a/packages/layers/src/events/LayerHandlers.ts
+++ b/packages/layers/src/events/LayerHandlers.ts
@@ -51,11 +51,7 @@ export class LayerHandlers extends DerivedCoreEventHandlers<{
 
             const { indicator, currentCanvasHovered } = LayerHandlers.events;
 
-            if (
-              currentCanvasHovered &&
-              indicator &&
-              currentCanvasHovered.data.nodes
-            ) {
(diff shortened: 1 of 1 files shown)


prevwong (approved):
Thank you! 💯

PR #564, opened 2023-09-29
fix: onNodesChange now updates its state correctly

akucinskii:
There is a problem with onNodesChange using state from the initialization of <Editor/> component 

This fixes it at now it updates the value correctly: 

Old code example: 

```ts
 console.log(mode) //1st call: firstMode, 2nd call: secondMode
  return (
    <Editor
      enabled={true}
      onNodesChange={(query) => {
        console.log(mode); //1st call: firstMode, 2nd call: firstMode (always firstMode no matter what)
      
      }}
      resolver={{ ... }}
    >
      {children}
    </Editor>
  );

```

New code example: 
```ts
 console.log(mode) //1st call: firstMode, 2nd call: secondMode
  return (
    <Editor
      enabled={true}
      onNodesChange={(query) => {
        console.log(mode); //1st call: firstMode, 2nd call: secondMode
      
      }}
      resolver={{ ... }}
    >
      {children}
    </Editor>
  );

```

Screenshot of working example: 

image

Code of this example screenshot:
 ```ts
const [count, setCount] = useState(0);

  return (
    <>
      <Editor
        resolver={{
          Item: Item,
          Toolbox: Toolbox,
        }}
        onNodesChange={() => {
          console.log('nodes changed, current count state: ', count);
        }}
      >
...
```

### DIFF
--- a/packages/core/src/editor/Editor.tsx
+++ b/packages/core/src/editor/Editor.tsx
@@ -76,6 +76,17 @@ export const Editor: React.FC<React.PropsWithChildren<Partial<Options>>> = ({
       return;
     }
 
+
+     if (
+      options.onNodesChange !== undefined &&
+      options.onNodesChange !== context.query.getOptions().onNodesChange
+    ) {
+      context.actions.setOptions((editorOptions) => {
+        editorOptions.onNodesChange = options.onNodesChange;
+      });
+    }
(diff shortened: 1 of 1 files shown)


Savuelo:
Same issue here please fix

PR #566, opened 2023-10-04
fix: ensure non-null resolver value in Editor

ismajl-ramadani:
In my case, deserializeNode throws an error because the resolver object is being passed as null. This mostly happens if resolvers are constructed in an async way, and then passed to the Editor as a null object because of a bug in the client side, the Editor will not warn for the incorrect value of the resolvers and then later behave strangely.


prevwong (approved):
Thanks! 💯

PR #568, opened 2023-10-13
chore(doc): add an example for overriding Editor event handlers

giliamverheide:
Added an example to the docs of how users can override the default Editor event handlers.
Note: Unlike the other examples, the code does include TypeScript types, let me know if you want the example to be JSX instead.

### DIFF
--- a/site/docs/api/Editor.md
+++ b/site/docs/api/Editor.md
@@ -116,9 +116,7 @@ import {
   NodeId
 } from '@craftjs/core'
 
-type CustomEventHandlersOptions = {};
-
-class CustomEventHandlers extends DefaultEventHandlers<CustomEventHandlersOptions> {
+class CustomEventHandlers extends DefaultEventHandlers {
   handlers() {
     const defaultHandlers = super.handlers()
 
@@ -152,7 +150,7 @@ const App = () => {
         new CustomEventHandlers({ store, isMultiSelectEnabled: () => false })
       }
     >
-      ..
+      ...
     </Editor>
   )
 }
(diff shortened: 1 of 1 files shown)


site/docs/api/Editor.md
```diff
@@ -100,4 +101,64 @@ const App = () => {
...
+  NodeId
+} from '@craftjs/core'
+
+class CustomEventHandlers extends DefaultEventHandlers {
+  isMultiSelectEnabled: (e: MouseEvent) => boolean
+
+  constructor(options: DefaultEventHandlersOptions & { store: EditorStore }) {
+    super(options)
+    this.isMultiSelectEnabled = options.isMultiSelectEnabled
+  }
```
prevwong, line 125: You can remove these lines. The `DefaultEventHandlers` constructor accepts an arbitrary `options` parameter in the constructor.

Although we should make an improvement to the `DefaultEventHandlers` class to accept a generic for options:
```tsx
export class DefaultEventHandlers<O = {}> extends CoreEventHandlers<
  DefaultEventHandlersOptions & O
>
```

Then in the custom event handler, we could do this:
```tsx
type CustomEventHandlersOptions = {}; 

class CustomEventHandlers extends DefaultEventHandlers<CustomEventHandlersOptions> {}
```

site/docs/api/Editor.md
```diff
@@ -100,4 +101,64 @@ const App = () => {
...
+
+  handlers() {
+    const defaultHandlers = super.handlers()
+
+    return {
+      ...defaultHandlers,
+      // Customize the hover event handler
+      hover: (el: HTMLElement, id: NodeId) => {
+        const unbindMouseover = defaultHandlers.hover(el, id)
```
prevwong, line 134: Perhaps it's better to name this variable `unbindDefaultHoverHandler` since the internal `hover` could technically contain events other than just `mouseover`

prevwong (approved):
Thank you! 💯

PR #569, opened 2023-10-13
feat: remove node hovered state on mouseleave event

giliamverheide:
Added `mouseleave` to default hover event handler, the `hovered` state gets removed when this is triggered.

Previously, if you wanted to removed the `hovered` state, you'd have to add something like
`ref={(ref: HTMLDivElement) => connectors.select(connectors.hover(ref, ''), '')}`
to a parent container to deselect/remove the hover state when the mouse pointer clicks/is outside of a node.

Also updated the `useEffect` in `packages/core/src/editor/Editor.tsx` due to lint error.

### DIFF
--- a/.changeset/rude-bees-deliver.md
+++ b/.changeset/rude-bees-deliver.md
@@ -2,4 +2,4 @@
 '@craftjs/core': patch
 ---
 
-Added mouseleave event to default hover event handler that removes hovered state when triggered
+Remove node hovered state on mouseleave event
--- a/.changeset/rude-bees-deliver.md
+++ b/.changeset/rude-bees-deliver.md
@@ -1,5 +1,5 @@
 ---
-'@craftjs/core': minor
+'@craftjs/core': patch
 ---
 
 Added mouseleave event to default hover event handler that removes hovered state when triggered
--- a/.changeset/rude-bees-deliver.md
+++ b/.changeset/rude-bees-deliver.md
@@ -1,0 +1,5 @@
+---
+'@craftjs/core': minor
+---
+
+Added mouseleave event to default hover event handler that removes hovered state when triggered
--- a/packages/core/src/editor/Editor.tsx
+++ b/packages/core/src/editor/Editor.tsx
@@ -79,7 +79,7 @@ export const Editor: React.FC<React.PropsWithChildren<Partial<Options>>> = ({
     }
 
     if (
-      options.enabled === undefined ||
+      options?.enabled === undefined ||
       context.query.getOptions().enabled === options.enabled
     ) {
       return;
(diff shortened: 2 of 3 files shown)


.changeset/rude-bees-deliver.md
```diff
@@ -0,0 +1,5 @@
+---
+'@craftjs/core': minor
```
prevwong, line 2: Should be `patch` 

https://github.com/prevwong/craft.js/blob/main/CONTRIBUTING.md#pull-requests

packages/core/src/editor/Editor.tsx
```diff
@@ -88,7 +88,7 @@ export const Editor: React.FC<React.PropsWithChildren<Partial<Options>>> = ({
     context.actions.setOptions((editorOptions) => {
       editorOptions.enabled = options.enabled;
     });
-  }, [context, options.enabled]);
+  }, [context, options?.enabled]);
```
prevwong, line 91: Let's keep the current change

examples/landing/next.config.js
```diff
@@ -1,3 +1,4 @@
 module.exports = {
-  assetPrefix: process.env.NODE_ENV === 'production' ? '/examples/landing' : '/',
+  assetPrefix:
+    process.env.NODE_ENV === 'production' ? '/examples/landing' : '/',
```
prevwong, line 3: Maybe something's up with your prettier config haha, let's keep the current change?

packages/core/src/events/DefaultEventHandlers.ts
```diff
@@ -134,8 +134,18 @@ export class DefaultEventHandlers<O = {}> extends CoreEventHandlers<
           }
         );
 
+        const unbindMouseleave = this.addCraftEventListener(
+          el,
+          'mouseleave',
+          (e) => {
+            e.craft.stopPropagation();
+            store.actions.setNodeEvent('hovered', '');
```
prevwong, line 142: Use `null` instead:

```suggestion
            store.actions.setNodeEvent('hovered', null);
```

packages/core/src/editor/Editor.tsx
giliamverheide: My linter wouldn't allow me to commit with the original `useEffect` code. Not sure if my eslint is wrong, but I could force commit the original code?

packages/core/src/events/DefaultEventHandlers.ts
prevwong: Yeah, Craft is using Typescript without strict mode thus this weird inconsistency. I will create a separate PR to migrate the repo to strict mode, but for now I've updated your PR to be consistent with the rest of the codebase.

prevwong (approved):
Thank you! 💯

PR #578, opened 2023-11-10
Allow adding custom metadata to the `UserComponent` that can be brought into `Node`

Traveller23:
Related discussion: https://github.com/prevwong/craft.js/discussions/576

After adding this feature, whenever we get the `Node`, we will also get the custom metadata of the user component. So we were able to automate some of the actions that we previously had to perform manually.
This can bring more configurability to the whole framework.

### DIFF
--- a/.changeset/stupid-onions-mate.md
+++ b/.changeset/stupid-onions-mate.md
@@ -1,5 +1,5 @@
 ---
-'@craftjs/core': minor
+'@craftjs/core': patch
 ---
 
 Add custom metadata to the UserComponent that can be brought into Node
--- a/.changeset/stupid-onions-mate.md
+++ b/.changeset/stupid-onions-mate.md
@@ -1,0 +1,5 @@
+---
+'@craftjs/core': minor
+---
+
+Add custom metadata to the UserComponent that can be brought into Node
--- a/packages/core/src/interfaces/nodes.ts
+++ b/packages/core/src/interfaces/nodes.ts
@@ -9,6 +9,7 @@ export type UserComponentConfig<T> = {
   related: Partial<NodeRelated>;
   props: Partial<T>;
   custom: Record<string, any>;
+  customMetadata: Record<string, any>;
   isCanvas: boolean;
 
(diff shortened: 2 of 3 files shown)


prevwong:
We can already achieve the same results with the `custom` prop already, no? 

```tsx
const Component = () => {}

Component.craft => {
  custom: {...}
} 
```

prevwong:
> > We can already achieve the same results with the `custom` prop already, no?
> > ```tsx
> > const Component = () => {}
> > 
> > Component.craft => {
> >   custom: {...}
> > } 
> > ```
> 
> There's a difference between `customMetadata` and `custom`: the data in `custom` is saved to the `SerializedNode` and thus to each instance. The data in `customMetadata` is shared by all instances of the component, so it doesn't need to be saved to `SerializedNode`.
> 
> Also, since we don't need to serialize `customMetadata`, we can put methods in there. We've been missing a means to attach methods to components before (and invoke those methods through `Node`), now we don't have to worry about that with `customMetadata`.

Ah gotcha. Can we rename `customMetadata` to something like `info`?

Traveller23:
> > > We can already achieve the same results with the `custom` prop already, no?
> > > ```tsx
> > > const Component = () => {}
> > > 
> > > Component.craft => {
> > >   custom: {...}
> > > } 
> > > ```
> > 
> > 
> > There's a difference between `customMetadata` and `custom`: the data in `custom` is saved to the `SerializedNode` and thus to each instance. The data in `customMetadata` is shared by all instances of the component, so it doesn't need to be saved to `SerializedNode`.
> > Also, since we don't need to serialize `customMetadata`, we can put methods in there. We've been missing a means to attach methods to components before (and invoke those methods through `Node`), now we don't have to worry about that with `customMetadata`.
> 
> Ah gotcha. Can we rename `customMetadata` to something like `info`?

Sure, I'll change it right away.

Traveller23:
@prevwong Done. If any other changes need to be made, just let me know.

prevwong (approved):
Thanks! ❤️

Issue #519, opened 2023-04-21
Drag shadows doesn't work on Safari


yuki-mo-zc:
**Describe the bug**
While dragging a component on Safari, the drag shadow doesn't show up. I tested this on both the landing and basic examples provided by craft js. It works fine on chrome. However, I found that this works properly on the new package [@craftjs/layers](https://github.com/prevwong/craft.js/tree/develop/packages/layers), while I am testing the landing example on Safari. 

**To Reproduce**

1. Open the basic / landing example on Safari 
2. drag a component 
3. the drag shadow doesn't show up while dragging the component
4. drag a component under layers in the landing example
5. the drag shadow does show up properly while dragging the component

**Expected behavior**
The drag shadow should show up properly while dragging the component

**Screenshots**
https://user-images.githubusercontent.com/23268712/233708377-d9177829-99ce-4b53-8871-2c68549a3986.mov

**Additional context**
I found this TODO in [this](https://github.com/prevwong/craft.js/blob/develop/packages/core/src/events/createShadow.ts#L1) file that seems to be related to this bug 
Since this issue is resolved in the new layers package, could we please apply the same fix to resolve it in the core package of craft.js? Thanks

### Your environment

| Software         | Version(s) |
| ---------------- | ---------- |
| craft.js         | ^0.2.0-beta.8
| React            | ^18.2.0
| TypeScript       | 4.9.4
| Browser          | Safari
| yarn         | 1.22.19
| Operating System | Mac OS

PR #587, opened 2023-12-01
fix: drag shadow on Safari

Accir:
Related to: https://github.com/prevwong/craft.js/issues/519

Fixes Safari drag shadow bug, where it is not visible when dragging. By changing display property to `absolute`, `background` is required as it will otherwise be transparent. By default I set the background to white. If any user wants to modify drag shadow css, I added additional classes `drag-shadow` and `drag-shadow-container`.

### DIFF
--- a/.changeset/shaggy-planets-camp.md
+++ b/.changeset/shaggy-planets-camp.md
@@ -1,0 +1,5 @@
+---
+'@craftjs/core': patch
+---
+
+fix Safari drag shadow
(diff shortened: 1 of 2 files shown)


Accir:
As I test this further I realised that `background` is not necessary as it is intended functionality that drag shadows have no background if the dragged element has no background. So only the position change should be sufficient to fix this bug.

prevwong (approved):
Thank you! ❤️

PR #607, opened 2024-01-26
fix: <Frame> docs bad json provided to data prop

noreplydev:
Fixes the Frame docs [Loading from serialized nodes](https://craft.js.org/docs/api/frame#loading-from-serialized-nodes) section. Now the example JSON provided in the data prop is bad formatted needs one more curly brace at the end. This curly brace is added on the PR, now the Frame docs example works well.


prevwong (approved):
Thanks! ❤️

PR #614, opened 2024-02-14
Allow external context injection to the Editor

noreplydev:
Now the @craftjs/core Editor component accepts an external context injection for advanced use cases when external management of the state is necessary. 

Previously, you needed to be inside the context provider to have access to the store. Now, since you can pass it as prop, you can initialise the context outside and have access to it. However, the component works as usual if you do not pass the context prop.


jcarlosn:
I also need this feature

theG3n15:
I need it too

PR #621, opened 2024-03-20
chore(docs): update type of state.events.selected in useNode hook

zzzhrookie:
In the useNode hook, both the selected and dragged are of type boolean, rather than array


prevwong (approved):
Thanks!

PR #623, opened 2024-03-25
fix: improve resolver performance

Traveller23:
After a certain number of components have been registered, rendering after deserialization becomes very slow, and this issue has been discussed many times on discord.
So I tried to solve this problem, and I finally realized that the root of the problem is the `resolveComponent` method. In the original implemention, the time complexity of the search was O(N^2 * LogN), so as soon as the number of components increased, the time needed to search increased at a faster rate.
I did two things: made the search time complexity O(LogN), and created a reverse index of the map and cached it.

This problem only occurs in the prodcution version, i.e. after the code has been compiled and converted to ES5 form. So I guess the problem is also related to the target javascript version.

### DIFF
--- a/packages/core/src/utils/resolveComponent.ts
+++ b/packages/core/src/utils/resolveComponent.ts
@@ -4,44 +4,63 @@ import invariant from 'tiny-invariant';
 
 import { Resolver } from '../interfaces';
 
-let currResolver: Resolver | undefined = undefined;
-const reversedResolver = new Map<React.ElementType | string, string>();
+type ReversedResolver = Map<React.ComponentType | string, string>;
 
-export const resolveComponent = (
-  resolver: Resolver,
-  comp: React.ElementType | string
-) => {
-  const componentName = (comp as any).name || (comp as any).displayName;
-  const resolvedName = resolver[componentName]
-    ? componentName
-    : searchComponentInResolver(resolver, comp);
+type CachedResolverData = {
+  resolver: Resolver;
+  reversed: ReversedResolver;
+};
 
-  invariant(
-    resolvedName,
-    ERROR_NOT_IN_RESOLVER.replace('%node_type%', componentName)
-  );
+let CACHED_RESOLVER_DATA: CachedResolverData | null = null;
 
-  return resolvedName;
+const getReversedResolver = (resolver: Resolver): ReversedResolver => {
+  if (CACHED_RESOLVER_DATA && CACHED_RESOLVER_DATA.resolver === resolver) {
+    return CACHED_RESOLVER_DATA.reversed;
+  }
+
+  CACHED_RESOLVER_DATA = {
+    resolver,
+    reversed: new Map(),
+  };
+
+  for (const [name, comp] of Object.entries(resolver)) {
+    CACHED_RESOLVER_DATA.reversed.set(comp, name);
+  }
+
+  return CACHED_RESOLVER_DATA.reversed;
+};
+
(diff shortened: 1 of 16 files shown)


packages/core/src/utils/resolveComponent.ts
```diff
@@ -1,34 +1,20 @@
 import { ERROR_NOT_IN_RESOLVER } from '@craftjs/utils';
+import React from 'react';
 import invariant from 'tiny-invariant';
 
 import { Resolver } from '../interfaces';
 
+let currResolver: Resolver | undefined = undefined;
+const reversedResolver = new Map<React.ElementType | string, string>();
+
 export const resolveComponent = (
```
prevwong, line 10: nit:
```tsx
type ReversedResolver = Map<React.ComponentType | string, string>;

type CachedResolverData = {
  resolver: Resolver; 
  reversed: ReversedResolver;
}

let CACHED_RESOLVER_DATA: CachedResolverData | null = null;

const getReversedResolver = (resolver: Resolver): ReversedResolver => {
  if ( CACHED_RESOLVER_DATA && CACHED_RESOLVER_DATA.resolver === resolver ) {
    return CACHED_RESOLVER_DATA.reversed;
  }

  CACHED_RESOLVER_DATA = {
    resolver,
    reversed: new Map();

  for ( const [name, comp] of Object.entries(resolver) {...}
}

const searchComponentInResolver = (
  resolver: Resolver,
  comp: React.ElementType | string
): string | null => {
  const name = getReversedResolver(resolver).get(comp);

   if (name !== null) {
     return name;
   }

  if (typeof comp === 'string') {
     return comp;
  }

  return null;
};

File: packages/core/src/utils/resolveComponent.ts [hunk 2]
```diff
@@ -1,34 +1,20 @@
...
-
-    if (typeof comp === 'string') {
-      return comp;
-    }
-  };
-
-  const resolvedName = getNameInResolver();
+  const resolvedName = resolver[componentName]
+    ? componentName
+    : searchComponentInResolver(resolver, comp);
```
prevwong, line 17: nit: Think we can just simplify this to `searchComponentInResolver` in all cases since it won't really impact efficiency after the initial call:

```tsx
const resolvedName = searchComponentInResolver(resolver, comp);
```

prevwong (commented):
Thanks, looks good. Just left some nits

Traveller23:
Okay, I'll change it in two days.

Traveller23:
@prevwong One question:
Is it possible to bring this code to the top?

```javascript
if (typeof comp === 'string') {
   return comp;
}
```

It would save a lot of searching time if we just know that the type of the `comp` is `string` and return it.

Traveller23:
I'm done refactoring, please check it again.
If the assumption above are incorrect, please let me know and I'll refactor the code again.

packages/core/src/utils/resolveComponent.ts
```diff
@@ -1,38 +1,65 @@
...
+  resolver: Resolver,
+  comp: React.ElementType
+): string | null => {
+  const componentName = getComponentName(comp);
+  if (Object.hasOwn(resolver, componentName)) {
+    return componentName;
+  }
+
+  const name = getReversedResolver(resolver).get(comp);
+  return name !== undefined ? name : null;
```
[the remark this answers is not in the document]
Traveller23: That's true, I'll remove those lines now.

Traveller23:
@prevwong  All done.

prevwong (approved):
Thank you! ❤️

Issue #200, opened 2021-02-24
Server side rendering possibility


timc1:
To support rendering Craft nodes into static markup on the server, we've had to hack around the way blocks are built – instead, it would be nice to just build various blocks the way they're supposed to be as per the docs and have SSR built-in.

What would enabling rendering `Element`s entail? Either way seems pretty critical to get this type of support for a landing page!

visualcookie:
> To support rendering Craft nodes into static markup on the server, we've had to hack around the way blocks are built – instead, it would be nice to just build various blocks the way they're supposed to be as per the docs and have SSR built-in.
> 
> What would enabling rendering `Element`s on the server entail? Either way seems pretty critical to get this type of support for a landing page!

Hej, I'm currently looking into SSR with craft.js and saw that you figured out a hacky way. Mind presenting how you solved it until this Feature gets really implemented into craft.js? Would be greatly appreciated.

hugominas:
I have also tried implementing this feature a couple of times but got stuck, the components were not rendered properly.

It would be outstanding to have option to pre-rendering the components on the server and let next.js handle tha static caching after. It would a great upgrade for SEO . I hope @prevwong has some time to help us with this.

desis123:
yes that will be great addition

mattvb91:
@timc1 could you share some code for your solution?

hennessyevan:
Has anyone in user-land accomplished this?

pepas24:
I've made a functional app with server side rendering using Craft and Next JS, based on a comment I read time ago (I don't remeber where but here in the issues or discussions).

Its a private repo, but the main idea is to have a wrapper component with all Craft related behavior and then replace them with naked components when are loaded by Next. I think is a good approach almost for my use case.

This is the important part

```jsx
const UserNode = ({ node, data }: any): any => {
  let typeName = ""

  if (typeof node.type === "object") {
    typeName = node.type.resolvedName
  } else {
    typeName = node.type
  }

  const Children = node.nodes.map((x: any) => {
    return <UserNode key={x} node={data[x]} data={data} />
  })

  switch (typeName) {
    case "ButtonCraft":  // use the sufix Craft convention to keep both components separated
      return <Button {...node.props} /> // then I switch to naked component with their props
    case "ContainerCraft":
      return <Container {...node.props}>{Children}</Container>
    default:
      return <></>
  }
}

export default function RenderLayout({ layout }: any) {
  // pass Craft layput to the recursive component
[...]

Issue #140, opened 2020-11-14
Add SSR Support


Robbilie:
**Is your feature request related to a problem? Please describe.**
I am trying to have a page rendered serverside. i have the serialized nodes there and want the page to prerender but it seems to editor/frame is doing it async.

**Describe the solution you'd like**
Add the abbility to render in a synchronous or blocking way (ie. a helper that outputs react elements based on serialized nodes).

dbousamra:
@Robbilie This may not be exactly what you want, but perhaps it's helpful https://github.com/prevwong/craft.js/issues/42#issuecomment-733255704

Robbilie:
ill look into it, a "native" option would be nice though :)

tabvn:
I think this feature should implemented for view mode. or we can add new render for view mode and one for editing mode.

PR #637, opened 2024-05-15
feat: SSR support

tomas-c:
Before these changes, the Frame and Element components delayed rendering user components until after the first render due to their reliance on useEffect. This meant that (a) server-side rendering did not work (b) pages with many deeply nested components took many render iterations to fully render a page which on slower devices would cause a visible layout shift.

After these changes neither Element nor Frame rely on useEffect.

Resolves:
- #140
- #200 
- #42

### DIFF
--- a/packages/core/src/render/Frame.tsx
+++ b/packages/core/src/render/Frame.tsx
@@ -44,21 +44,21 @@ export const Frame: React.FC<React.PropsWithChildren<FrameProps>> = ({
     initialData: data || json,
   });
 
-  const isInitialChildrenLoadedRef = useRef(false);
+  const isInitialLoadedRef = useRef(false);
+
+  if (!isInitialLoadedRef.current && initialState.current.initialData) {
+    actions.history.ignore().deserialize(initialState.current.initialData);
+    isInitialLoadedRef.current = true;
+  }
 
   useEffect(() => {
-    const { initialChildren, initialData } = initialState.current;
-
-    if (initialData) {
-      actions.history.ignore().deserialize(initialData);
-      return;
-    }
+    const { initialChildren } = initialState.current;
 
     // Prevent recreating Nodes from child elements if we already did it the first time
     // Usually an issue in React Strict Mode where this hook is called twice which results in orphaned Nodes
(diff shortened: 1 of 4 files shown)


hugominas:
@tomas-c great work, this is such a needed feature @prevwong would be great to have your eyes on this PR

packages/core/src/render/Frame.tsx
```diff
@@ -44,21 +44,21 @@ export const Frame: React.FC<React.PropsWithChildren<FrameProps>> = ({
...
+    actions.history.ignore().deserialize(initialState.current.initialData);
+    isInitialLoadedRef.current = true;
+  }
 
-    if (initialData) {
-      actions.history.ignore().deserialize(initialData);
-      return;
-    }
+  useEffect(() => {
+    const { initialChildren } = initialState.current;
```
prevwong, line 55: I think you we can simplify this to:
```tsx
  const isLoaded = useRef(false);

  if (!isLoaded.current) {
    const initialData = data || json;

    if (initialData) {
      actions.history.ignore().deserialize(initialData);
    } else {
      const rootNode = React.Children.only(children) as React.ReactElement;

      const node = query.parseReactElement(rootNode).toNodeTree((node, jsx) => {
        if (jsx === rootNode) {
          node.id = ROOT_NODE;
        }
        return node;
      });

      actions.history.ignore().addNodeTree(node);
    }
```

so we can remove the `useEffect` hook completely

prevwong (commented):
Looks good, just left a suggestion

packages/core/src/render/Frame.tsx
tomas-c: Thanks! I've applied the changes. This was where I wasn't sure if Creaft.js somehow relied on the children parsing being done after rendering. As far as I can tell looking in more detail - it doesn't. And tests do pass and I couldn't see anything breaking while playing around with the examples in the repo and my own app. So hopefully good to go!

prevwong: Yeap, Craft just parses the children elements if there was no `data` value present - so in this way, the children elements are used to create the default/initial content for the editor

prevwong (approved):
Thank you! 💯

PR #656, opened 2024-06-03
chore: export dnd utils

mateusz-oi:
Hello,

I would like to suggest adding exports for Positioner and createShadow. This enhancement would provide more possibilities for handler customization, as outlined in the [Craft.js documentation on overriding default event handlers](https://craft.js.org/docs/api/editor#override-default-event-handlers).

We need this to be able to partially use defaultHandlers and modify the create method, leveraging default methods like Positioner and incorporating effects such as createShadow during drag events.

Thank you for considering this suggestion.


prevwong (approved):

prevwong:
Thanks! 💯

PR #705, opened 2024-09-27
 feat: allow layers to utilise CoreEventHandler mouseleave options

itsbjoern:
I noticed that `removeHoverOnMouseleave` wasn't working in the Layers, to get this working I basically just piggy-backed on the same option that is used in the core package and applied the relevant listener.

```ts
if (this.derived.options.removeHoverOnMouseleave) {
  unbindMouseleave = this.addCraftEventListener(
    el,
    'mouseleave',
    (e) => {
      e.craft.stopPropagation();
      layerStore.actions.setLayerEvent('hovered', null);
    }
  );
}
```

Not 100% sure on the change of adding the option to the type of the [CoreEventHandlers]( https://github.com/itsbjoern/craft.js/blob/4a20cff03a911a062a8bdf3fa523672bcc450d61/packages/core/src/events/CoreEventHandlers.ts#L11) but as far as I can tell that resolves any type issues on the derived handler type.

Also, as part of it I snuck in a small fix, as far as I can tell the mouseleave listener was never cleaned up in core because it was actually never assigned, so I added this as part of this PR.

### DIFF
--- a/.changeset/smooth-elephants-clean.md
+++ b/.changeset/smooth-elephants-clean.md
@@ -1,0 +1,6 @@
+---
+'@craftjs/core': patch
+'@craftjs/layers': patch
+---
+
+Use removeHoverOnMouseleave option for layers package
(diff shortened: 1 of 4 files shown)


prevwong (approved):
Nice work, thank you! 🚀

Issue #698, opened 2024-09-07
Error involving layers in nextjs


RoyMtonga:
when i use the layers component in my project it qorks but i get erros in my cosole..
app-index.tsx:25 
 Warning: React does not recognize the `isHidden` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `ishidden` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
app-index.tsx:25 
 Warning: Received `false` for a non-boolean attribute `expanded`.

If you want to write it to the DOM, pass a string instead: expanded="false" or expanded={value.toString()}.

If you used to conditionally omit it with expanded={condition && value}, pass expanded={condition ? value : undefined} instead.
    at a
app-index.tsx:25 
 Warning: React does not recognize the `hasCanvases` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `hascanvases` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
app-index.tsx:25 
 Warning: Received `false` for a non-boolean attribute `hovered`.

If you want to write it to the DOM, pass a string instead: hovered="false" or hovered={value.toString()}.

If you used to conditionally omit it with hovered={condition && value}, pass hovered={condition ? value : undefined} instead.

Issue #672, opened 2024-07-11
Decouple @craftjs/layers with styled-components


addlistener:
My project does not depend on `styled-components`, however @craftjs/layers will depend on it even tho I provide a custom render. The same thing applies to `react-contenteditable`.

https://github.com/prevwong/craft.js/blob/46eb68ef54a036f06c9433ee74b2359a8f38c610/packages/layers/src/layers/DefaultLayer/DefaultLayerHeader.tsx#L3

Ideally the `layers` should be headless? I guess we can move the default layer component inside the examples?

addlistener:
Plus there are some warnings here because of the UI. A bit annoying lol

<img width="1177" alt="image" src="https://github.com/prevwong/craft.js/assets/4531670/e022b6c8-3af4-4f40-a65a-9ce721169336">

<img width="1180" alt="image" src="https://github.com/prevwong/craft.js/assets/4531670/e46d30df-fdf0-4bc0-b0e2-b9bcf4b3e569">


<img width="1186" alt="image" src="https://github.com/prevwong/craft.js/assets/4531670/bb660394-46f2-4695-a223-e6a0234f5930">

<img width="1186" alt="image" src="https://github.com/prevwong/craft.js/assets/4531670/e71c3809-0449-49f7-8efe-e88934564a61">

PR #707, opened 2024-10-16
fix: Update styled-components

pixel-toys-chris-evans:
Fixes: #698 and #672 (Comments made below initial issue posting, and not a direct fix of the main topic).

Anyone working with a newer version of StyledComponents is unable to run CraftJS Layers in production due to the errors raised by breaking changes introduced in `"styled-components": ">= 6"`. 

We faced this while attempting to use CraftJS within Sanity CMS.

### DIFF
--- a/examples/landing/components/editor/RenderNode.tsx
+++ b/examples/landing/components/editor/RenderNode.tsx
@@ -2,7 +2,7 @@ import { useNode, useEditor } from '@craftjs/core';
 import { ROOT_NODE } from '@craftjs/utils';
 import React, { useEffect, useRef, useCallback } from 'react';
 import ReactDOM from 'react-dom';
-import styled from 'styled-components';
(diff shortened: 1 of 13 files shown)


.changeset/red-cameras-burn.md
```diff
@@ -0,0 +1,6 @@
+---
+'@craftjs/layers': minor
+'@craftjs/core': patch
+---
```
prevwong, line 4: 1. I don't think the core package needs an update since the changes here only affect the layers package?
2. Can you change the `@craftjs/layers` as a `patch` update instead? It's not ideal but the reason being we're still in major version 0 and we use the minor version to indicate a breaking change in Craft

prevwong (commented):
Thanks, looks good! Just left a note on the changeset

prevwong (approved):
Thank you! 🚀

PR #732, opened 2025-03-09
fix: Deprecated: element.ref

manelpb:
Accessing element.ref is no longer supported. ref is now a regular prop. 

Update to address the issue.

### DIFF
--- a/packages/utils/src/EventHandlers/wrapConnectorHooks.tsx
+++ b/packages/utils/src/EventHandlers/wrapConnectorHooks.tsx
@@ -19,7 +19,7 @@ export function cloneWithRef(
   element: any,
   newRef: any
 ): React.ReactElement<any> {
-  const previousRef = element.props.ref;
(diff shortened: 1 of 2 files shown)


packages/utils/src/EventHandlers/wrapConnectorHooks.tsx
```diff
@@ -19,7 +19,7 @@ export function cloneWithRef(
   element: any,
   newRef: any
 ): React.ReactElement<any> {
-  const previousRef = element.ref;
+  const previousRef = element.props.ref;
```
prevwong: Wondering if this would break on older versions of React though? Or would something like `element.props.ref || element.ref` make sense for backwards compatibility?

manelpb: Yeah I could try adding for both

ddudev (approved):

